PR #1, opened 2021-03-11
add postgresql for prod

codiebeulaine:


contentrepo/settings/production.py
```diff
@@ -6,3 +6,20 @@
...
+DATABASES = {
+    'default': {
+        'ENGINE': 'django.db.backends.postgresql_psycopg2',
+        'NAME': os.environ.get('DATABASE_NAME','name'),
+        'USER': os.environ.get('DATABASE_LOGIN','user'),
+        'PASSWORD': os.environ.get('DATABASE_PASSWORD', 'password'),
+        'HOST': os.environ.get('DATABASE_HOST', ''),
+        'PORT': '5432',  # Set to empty string for default.
+        'CONN_MAX_AGE': 600,
+    }
```
KaitCrawford, line 19: I think we usually do this by installing dj_database_url and then doing it like https://github.com/praekeltfoundation/casepro-docker/blob/master/settings.py#L27-L31
but this is fine for now

contentrepo/settings/production.py
```diff
@@ -6,3 +6,20 @@
...
+}
+
+SECRET_KEY = os.environ.get(
+    'SECRET_KEY', 'vg%%_7^bt(68qg50vgp^*f^&_(f)36k1vnyg%49n8*0s86hxj6')
```
KaitCrawford, line 23: You've committed a secret key here. you should put a placeholder here and revoke this key

KaitCrawford (approved):
:+1:

PR #3, opened 2021-04-12
Add wagtail content import

codiebeulaine:
Add wagtail content import. It creates a button on a page that allows the option to import content from a google doc and maps it to our web content. This is only for web content


KaitCrawford (approved):
:+1:

PR #7, opened 2021-04-19
Import content via JSON

codiebeulaine:
## Purpose
_Import content from one CMS to another_

## Approach
_The quickest and easiest way for us to do this is to add a script to existing project that exports desired fields to json, and then run a management command to import it on the content repo_

### DIFF
--- a/home/tests/output.json
+++ b/home/tests/output.json
@@ -1,0 +1,1 @@
+{"articles": [["article title", "article subtitle", ["<p>this is some plain text</p>", "<p>this is some richtext</p>"]]]}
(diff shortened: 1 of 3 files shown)


home/tests/output.json
```diff
@@ -0,0 +1 @@
+{"articles": [["article title", "article subtitle", ["<p>this is some plain text</p>", "<p>this is some richtext</p>"], ["tag1", "tag2"]]]}
```
KaitCrawford, line 1: shouldn't each element of the articles list be a dict, rather than a list?
Then ordering of the title, subtitle, ect won't matter

KaitCrawford (commented):
question about the articles being lists instead of dicts

home/tests/output.json
codiebeulaine: will change

KaitCrawford (approved):
:+1:

PR #9, opened 2021-04-21
Add platform char limits

codiebeulaine:
## Purpose
_Our different platforms have different message limits, we need t limit our content length accordingly_

## Approach
_Limit content before it gets to the DB_

- [x] Add help text so admins know the limit
- [x] Add validation method for limits

## Learning
_Wagtail does weird stuff when it comes to adding individual errors to streamfields. We can't put the errors in the field themselves and have to add it to the admin form_

I encountered [this](https://github.com/wagtail/wagtail/issues/5663) error


home/models.py
```diff
@@ -15,6 +15,34 @@
...
+        for block in messenger_body:
+            if block.block_type == "paragraph":
+                total_count += len(block.render())
+        if total_count >= 2000:
+            self.add_error(None, 'Messenger body exceeds 2000 characters')
+        for block in viber_body:
+            if block.block_type == "paragraph":
+                total_count += len(block.render())
+        if total_count >= 2000:
+            self.add_error(None, 'Viber body exceeds 7000 characters')
```
KaitCrawford, line 44: I think either the error message or the count check here is in correct? the message says 7000 but the check is for 2000

KaitCrawford (approved):
Comment about the Viber limit check but :+1: when that's resolved

PR #10, opened 2021-05-19
Security Check

codiebeulaine:
## Purpose
_A secret key was added while we figured out the correct configuration for our QA. Going forward, before creating any more instances, we would like to get the secret key from env vars as this repo is public_

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,9 +30,8 @@ COPY --chown=wagtail:wagtail . .
 
 USER wagtail
 
-ENV DJANGO_SETTINGS_MODULE=contentrepo.settings.base
+ENV DJANGO_SETTINGS_MODULE=contentrepo.settings.production
 
-RUN SECRET_KEY='vg%%_7^bt(68qg50vgp^*f^&_(f)36k1vnyg%49n8*0s86hxj6' python manage.py collectstatic --noinput --clear
-RUN SECRET_KEY='vg%%_7^bt(68qg50vgp^*f^&_(f)36k1vnyg%49n8*0s86hxj6' python manage.py migrate
+RUN python manage.py collectstatic --noinput --clear
 
 CMD set -xe; gunicorn contentrepo.wsgi:application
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -30,9 +30,8 @@ COPY --chown=wagtail:wagtail . .
 
 USER wagtail
 
-ENV DJANGO_SETTINGS_MODULE=contentrepo.settings.base
+ENV DJANGO_SETTINGS_MODULE=contentrepo.settings.production
 
-RUN SECRET_KEY='vg%%_7^bt(68qg50vgp^*f^&_(f)36k1vnyg%49n8*0s86hxj6' python manage.py collectstatic --noinput --clear
-RUN SECRET_KEY='vg%%_7^bt(68qg50vgp^*f^&_(f)36k1vnyg%49n8*0s86hxj6' python manage.py migrate
```
KaitCrawford, line 36: I recommend rotating these keys. I don't see any place where the codebase is copied to the docker image but if it is then the git history will also be copied, including this commit that removes the keys.

contentrepo/settings/base.py
```diff
@@ -3,8 +3,10 @@
 DEBUG = True
 PROJECT_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 BASE_DIR = os.path.dirname(PROJECT_DIR)
-SECRET_KEY = 'vg%%_7^bt(68qg50vgp^*f^&_(f)36k1vnyg%49n8*0s86hxj6'
-
+SECRET_KEY = os.environ.get(
+    'SECRET_KEY', 'vg%%_7^bt(68qg50vgp^*f^&_(f)36k1vnyg%49n8*0s86hxj6')
```
KaitCrawford, line 7: I think it's better to not have a default Secret_key. that way if we forget to put one in the Env vars the deploy will fail, rather than the application quietly running with a vulnerable key.

Dockerfile
codiebeulaine: okay, rotating

contentrepo/settings/base.py
codiebeulaine: okay

KaitCrawford (approved):

PR #11, opened 2021-05-19
Add pagination tests

codiebeulaine:
## Purpose
_Creating tests for message pagination and fix flake8_

### DIFF
--- a/contentrepo/settings/base.py
+++ b/contentrepo/settings/base.py
@@ -128,7 +128,7 @@ STATICFILES_DIRS = [
 ]
 
 
-STATICFILES_STORAGE = 'django.contrib.staticfiles.storage.ManifestStaticFilesStorage'
+STATICFILES_STORAGE = 'django.contrib.staticfiles.storage.ManifestStaticFilesStorage'  # noqa
 
 STATIC_ROOT = os.path.join(BASE_DIR, 'static')
 STATIC_URL = '/static/'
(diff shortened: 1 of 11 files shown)


contentrepo/settings/base.py
```diff
@@ -145,8 +147,10 @@
     ('es', "Spanish"),
 ]
 
-# Base URL to use when referring to full URLs within the Wagtail admin backend -
-# e.g. in notification emails. Don't include '/admin' or a trailing slash
+# Base URL to use when referring to full URLs within the
+# Wagtail admin backend -
+# e.g. in notification emails. Don't include '/admin'
+# or a trailing slash
 BASE_URL = 'http://example.com'
```
KaitCrawford, line 154: This should probably be an actual url or pull from env vars

home/tests/test_api.py
```diff
@@ -0,0 +1,61 @@
...
+        content = json.loads(response.content)
+        self.assertEquals(
+            content, {'detail': 'The requested message does not exist'})
+
+        # it should return an appropriate error if requested message is not
+        # a positive integer value
+        response = self.client.get(
+            "/api/v2/pages/4/?whatsapp=True&message=notint")
+        content = json.loads(response.content)
+        self.assertEquals(
```
KaitCrawford, line 57: does this return a http 400 status code?

KaitCrawford (approved):
:+1: minor comments/questions that don't have to affect this PR

I think it would also be good to return the number of messages for the whole page. so we can display "page 3 of 6" or whatever

home/tests/test_api.py
codiebeulaine: it does now

KaitCrawford (approved):
:+1: :star2: :cake:

PR #13, opened 2021-05-24
add tag filtering

codiebeulaine:
## Purpose
_Need to be able to filter pages via tag_

### DIFF
--- a/home/api.py
+++ b/home/api.py
@@ -3,7 +3,6 @@ from wagtail.api.v2.router import WagtailAPIRouter
 from wagtail.images.api.v2.views import ImagesAPIViewSet
 from wagtail.documents.api.v2.views import DocumentsAPIViewSet
 from .serializers import ContentPageSerializer
-from django.views.decorators.cache import cache_page
 from.models import ContentPageTag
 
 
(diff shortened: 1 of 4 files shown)


home/api.py
```diff
@@ -3,19 +3,29 @@
...
+    ])
+
+    def get_queryset(self):
+        queryset = super(ContentPagesViewSet, self).get_queryset()
+        tag = self.request.query_params.get('tag')
+        if tag is not None:
+            ids = []
+            for t in ContentPageTag.objects.all():
+                if t.tag.name == tag:
+                    ids.append(t.content_object_id)
```
KaitCrawford, line 22: Is this a list of Tag objects?
Is that what we're returning or are we returning content pages that have these tags?

codiebeulaine: we are returning the content pages. so we collect the IDs fo the pages that match the tag then filter them against the queryset given by super

KaitCrawford: ah I see. `t.content_object_id` is what I missed. so this is a list of content page ids

KaitCrawford (approved):
:+1:

PR #14, opened 2021-05-26
add page number pagination

codiebeulaine:
## Purpose
_Default pagination for wagtail API is limit offset, but we prefer page number pagnination_


KaitCrawford (approved):
:+1:

PR #31, opened 2021-10-13
Add content import button

codiebeulaine:
## Purpose
_Add import button to import content from the CMS_


home/utils.py
```diff
@@ -0,0 +1,84 @@
...
+        else:
+            block = blocks.StructBlock(
+                [
+                    ("message", blocks.TextBlock()),
+                ]
+            )
+            block_value = block.to_python({"message": message["answer"]})
+            return block_value
+
+    # def create_tags(tags, page):
```
erikh360, line 46: can probably remove this

home/utils.py
```diff
@@ -0,0 +1,84 @@
...
+    #     for tag in tags:
+    #         created_tag, _ = Tag.objects.get_or_create(name=tag)
+    #         page.tags.add(created_tag)
+
+    ContentPage.objects.filter(id__lte=1000).delete()
```
erikh360, line 51: Why are we doing this in 3 lines?

home/utils.py
```diff
@@ -0,0 +1,84 @@
...
+    ContentPage.objects.filter(id__lte=2000).delete()
+    ContentPage.objects.filter(id__lte=3000).delete()
+    data = json.load(file)
+
+    for message in data["data"]:
+        try:
+            print(message)
```
erikh360, line 58: print

home/views.py
```diff
@@ -0,0 +1,22 @@
...
+from django.shortcuts import render
+from .forms import UploadFileForm
+from .utils import import_content
+
+
+def upload_file(request):
+    if request.method == "POST":
+        form = UploadFileForm(request.POST, request.FILES)
+        if form.is_valid():
+            job = handle_uploaded_file(request.FILES["file"])
```
erikh360, line 11: Can't we just call `import_content` directly here?

home/utils.py
codiebeulaine: wagtail delete only allows 1000 at a time

PR #34, opened 2021-12-15
Save page views

erikh360:


home/api.py
```diff
@@ -28,6 +34,14 @@ class ContentPagesViewSet(PagesAPIViewSet):
...
         super(ContentPagesViewSet, self).get(self, request, *args, **kwargs)
 
+    def detail_view(self, request, pk):
+        ContentPage.objects.get(id=pk).save_page_view(request.query_params)
+        return self._detail_view(request, pk)
+
+    @method_decorator(cache_page(60 * 60 * 2))
+    def _detail_view(self, request, pk):
+        return super().detail_view(request, pk)
+
```
erikh360, line 44: I did it this way to log a page view even if returns a cached page.

home/api.py
```diff
@@ -28,6 +28,10 @@ class ContentPagesViewSet(PagesAPIViewSet):
     def get(self, request, *args, **kwargs):
         super(ContentPagesViewSet, self).get(self, request, *args, **kwargs)
 
+    def detail_view(self, request, pk):
```
erikh360, line 31: The caching was messing with the tests so I removed it

home/models.py
```diff
@@ -286,3 +300,14 @@ class ContentPageRating(models.Model):
     helpful = models.BooleanField()
     comment = models.TextField(blank=True, default="")
     data = models.JSONField(default=dict, blank=True, null=True)
+
+
+class PageView(models.Model):
+    timestamp = models.DateTimeField(auto_now=True)
```
codiebeulaine, line 306: isn't it auto_now_add?

codiebeulaine (approved):

home/models.py
erikh360: yup

PR #35, opened 2021-12-24
Add list endpoints for page ratings and views

erikh360:


Buhle79 (approved):

home/tests/test_views.py
```diff
@@ -0,0 +1,158 @@
...
+
+        self.assertEquals(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEquals(
+            response.json(),
+            {
+                "page": ["Page matching query does not exist."],
+            },
+        )
+
+    def test_get_list(self):
```
KaitCrawford, line 80: This test and the one below would still pass if the view was incorrectly only returning 1 object. Perhaps you could add a 3rd object so 2 are returned, or make a separate test to confirm the querystring filtering is working?

KaitCrawford (approved):
Looks good
Just a small suggestion in the tests. :+1: either way

KaitCrawford:
Yeah :+1: looks better

PR #36, opened 2022-01-20
Content import with a newline (message break) argument

megh-k:
During the uploading of the OOY content, the only options for split messages was yes which splits each line into another message with `splitlines()`, or no, which results in only one message. We needed the OOY content to break in certain places to stop WhatsApp giving a "read more" link. 

I added an argument called newlines to the `import_csv_content` which breaks WhatsApp content into new messages (As I type this I realise my naming was terrible and it should probably change, maybe `messagebreak`?. Suggestions are welcomed). This allows me to set them to break into new messages on the custom message break `====`. I will also add a test if these changes are accepted. 

If there is another way to do this, please let me know. 
I also wasn't sure if this functionality was necessary for web pages.

### DIFF
--- a/home/management/commands/import_csv_content.py
+++ b/home/management/commands/import_csv_content.py
@@ -1,5 +1,4 @@
 import csv
-from email.policy import default
 from wagtail.core import blocks
 from django.core.management.base import BaseCommand
 from home.models import ContentPage, HomePage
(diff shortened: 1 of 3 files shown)


codiebeulaine:
I would add a tiny test for it

codiebeulaine (approved):

PR #40, opened 2022-01-28
Fix boto storage backend used

KaitCrawford:
## Purpose
S3BotoStorage is no longer available

## Approach
Use S3Boto3Storage instead


megh-k:
👍

PR #49, opened 2022-05-24
Feature/update csv upload add messenger and viber

megh-k:
Adds web, whatsapp, messenger, and viber content based on the headings in the CSV


home/management/commands/import_csv_content.py
```diff
@@ -86,16 +156,14 @@ def clean_row(row):
...
-                contentpage.save_revision().publish()
+                contentpage = add_web(row)
+                contentpage = add_whatsapp(row, contentpage)
+                contentpage = add_messenger(row, contentpage)
+                contentpage = add_viber(row, contentpage)
+
+                if contentpage:
+                    create_tags(row, contentpage)
+                    add_parent(row, contentpage, home_page)
+                    contentpage.save_revision().publish()
```
KaitCrawford, line 167: You might want to log something if a content page hasn't been created at this point

KaitCrawford (approved):
:+1: :cake: 
Although you might want to think about where else some logging could come in handy

home/utils.py
```diff
@@ -77,34 +80,109 @@ def clean_row(row):
             "web_body",
             "whatsapp_title",
             "whatsapp_body",
-            "parent",
+            "viber_title",
+            "viber_body" "messenger_title",
+            "messenger_body" "parent",
```
KaitCrawford, line 85: I think you're missing some commas on this line and the one above?

KaitCrawford (approved):
Small typo (I think), otherwise :+1:

home/utils.py
megh-k: weird, I deeeefinitely fixed this, I will fix it again quick

PR #54, opened 2022-07-19
Upgrade Django 3 to 4.0

codiebeulaine:
## Purpose
_Updating Django from 3 to 4.0 so that we can upgrade wagtail to version 3_

#### Blog Posts
- [Django 4.0 release notes](https://docs.djangoproject.com/en/4.0/releases/4.0/)

### DIFF
--- a/contentrepo/settings/dev.py
+++ b/contentrepo/settings/dev.py
@@ -7,7 +7,12 @@ DEBUG = True
 SECRET_KEY = os.environ.get("SECRET_KEY", DEFAULT_SECRET_KEY)
 
 # SECURITY WARNING: define the correct hosts in production!
-ALLOWED_HOSTS = os.environ.get("ALLOWED_HOSTS", "localhost")
+ALLOWED_HOSTS = os.environ.get(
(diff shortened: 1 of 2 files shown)


megh-k (approved):

contentrepo/settings/dev.py
```diff
@@ -7,7 +7,12 @@
 SECRET_KEY = os.environ.get("SECRET_KEY", DEFAULT_SECRET_KEY)
 
 # SECURITY WARNING: define the correct hosts in production!
-ALLOWED_HOSTS = os.environ.get("ALLOWED_HOSTS", "localhost")
+ALLOWED_HOSTS = os.environ.get(
+    "ALLOWED_HOSTS",
+    [
+        "localhost",
+    ],
+)
```
rudigiesler, line 15: I don't think this will work very well, since if we every define `ALLOWED_HOSTS`, then we'll get a string again instead of a list. We should probably just use `env.list` like in the production config, but with the `localhost` default

rudigiesler (commented):
Just a small comment on the settings change, otherwise looks good

PR #58, opened 2022-07-21
Add platform to the pageview model

megh-k:


home/models.py
```diff
@@ -365,6 +375,12 @@ class ContentPageRating(models.Model):
 
 
 class PageView(models.Model):
+    platform = models.TextField(
```
codiebeulaine, line 378: can you add choices here?

home/models.py
```diff
@@ -365,6 +375,18 @@ class ContentPageRating(models.Model):
 
 
 class PageView(models.Model):
+    platform = models.CharField(
+        choices=[
+            ("WHATSAPP", "whatsapp"),
+            ("VIBER", "viber"),
+            ("FACEBOOK", "facebook"),
```
codiebeulaine, line 382: can we change it from facebook to messenger

codiebeulaine: otherwise, 👍

codiebeulaine (approved):

PR #63, opened 2022-07-27
Add stale content report

codiebeulaine:
## Purpose
_Add report for stale content that is filterable by view count and last published_

#### TODOs in next PR
- [ ] Add platform to view count filter and list


erikh360 (approved):

rudigiesler (approved):
👍 Looks good. My only concern would be around performance and doing n+1 queries, although fixing that doesn't look super straightforward (I'm not sure if you can select_related for aggregates like count, maybe you can annotate and reference the annotated field in the view), so happy for that to happen at a later stage if we run into performance issues.

PR #75, opened 2022-08-22
Feat/export formatted content xlsx

megh-k:


home/utils.py
```diff
@@ -217,86 +248,338 @@ def add_viber(row, page=None):
...
+    for msg in platform_body:
+        msgs.append(
+            {
+                "msg": msg.value["message"] if "message" in msg.value else None,
+                "img": msg.value["image"] if "image" in msg.value else None,
+                "doc": msg.value["document"] if "document" in msg.value else None,
+                "med": msg.value["media"] if "media" in msg.value else None,
+                "next_prompt": msg.value["next_prompt"]
+                if "next_prompt" in msg.value
+                else None,
```
rudigiesler, line 263: Could also be `msg.get("next_prompt")`, or `msg.get("next_prompt", default=None)` if you want to be more explicit. But works either way. (assuming that this `msg` thing acts like a dictionary)

home/utils.py
```diff
@@ -1,16 +1,47 @@
...
+    "quick_replies",
+    "triggers",
+    "locale",
+    "next_prompt",
+    "image_link",
+    "doc_link",
+    "media_link",
+    "related_pages",
+]
+export_filename = f'exported_content_{datetime.now().strftime("%Y%m%d")}'
```
rudigiesler, line 43: The convention for these kind of global variables is screaming snake case, ie. `EXPORT_FILENAME`

home/utils.py
```diff
@@ -217,86 +248,338 @@ def add_viber(row, page=None):
...
+        if content_page.has_children:
+            add_children(temp_sheet, content_page.get_children())
+    return temp_sheet
+
+
+def get_content_sheet() -> List[list]:
+    content_sheet = []
+    headings = [1, 2, 3, 4, 5, "Message"] + fieldnames
+    content_sheet.append(headings)
+    home = HomePage.objects.filter(locale_id=1).first()
```
rudigiesler, line 549: I don't know wagtail well enough for this, but is there not a kind of home local, or default or root locale, that we can rely on instead of hardcoding the primary key here? The issue with something like this is if you add a row, then delete it, then add a row again, then that key doesn't get recycled, so you won't ever get a row with the ID of 1 again.

home/utils.py
```diff
@@ -217,86 +248,338 @@ def add_viber(row, page=None):
...
+        format_list_from_query_set(page.tags.all()),
+        format_list_from_query_set(page.quick_replies.all()),
+        format_list_from_query_set(page.triggers.all()),
+        str(page.locale),
+        get_next_prompts(messenging_platform_messages, 0),
+        get_image_link(messenging_platform_messages, 0),
+        get_doc_link(messenging_platform_messages, 0),
+        get_media_link(messenging_platform_messages, 0),
+        format_list_from_query_set(get_related_pages(page)),
+    ]
```
rudigiesler, line 505: Not for this PR, but maybe a change in the future, this section is the most confusing for me. There's nothing tying the arbitrary column positions with what they actually mean.

I'm wondering if using something like a dataclass here to hold all the data, might help better split the "getting the data" sections from the "formatting the data for presentation" sections.

eg.
```
@dataclass
class ExportRow:
    title: str
    subtitle: str
    body: str
```
Then once you've got all the rows, you can refer to things like `row.messenger_title`, which better describes what that actually is.

home/utils.py
```diff
@@ -217,86 +248,338 @@ def add_viber(row, page=None):
...
+
+
+def get_rows(page: ContentPage) -> List[list]:
+    """Sets up row for each page including the side panel.
+    Each page is returned as a list of rows"""
+    actual_depth = (
+        page.depth - 3
+    )  # main menu is depth 3 (is this always true? root -> home -> main menu)
+    padding = [""] * 6
+    padding[actual_depth] = "x"
```
rudigiesler, line 466: I'm wondering if there's a better way of doing this, rather than setting `x`'s and then overwriting them with the correct value later, couldn't we set them to the correct value here? (yet again, that looks like a big change, so probably not for this PR).

home/utils.py
```diff
@@ -217,86 +248,338 @@ def add_viber(row, page=None):
...
+    This will take the one found, can be extended to a list of urls
+    """
+    next_prompts = []
+    for platform_messages in messenging_platform_messages.values():
+        if platform_messages and len(platform_messages) < index - 1:
+            next_prompts.append(platform_messages[index]["next_prompt"])
+
+    next_prompt = next(
+        (next_prompt for next_prompt in next_prompts if next_prompt is not None), None
+    )
```
rudigiesler, line 358: This is quite confusing to me, it could probably be simplified by just returning out of the function when we find the first match (and then we also don't need to waste finding all the matches), ie:
```
for platform_messages in messenging_platform_messages.values():
        if (
            platform_messages
            and len(platform_messages) < index - 1
            and platform_messages[index]["next_prompt"] is not None
        ):
            return platform_messages[index]["next_prompt"]
return ""
 ```
 
 (and similar for where this pattern is repeated)

rudigiesler (commented):
I think there are a few small things that can be changed to make it more easily understood, and then a few bigger suggestions which are probably best left as separate PRs (if capacity allows). But overall, looks good :)

home/utils.py
megh-k: Unfortunately there is no get method :( I initially did try and eventually found this 

https://groups.google.com/g/wagtail/c/a6tA-ftoYD4/m/7s3Pm9gDBwAJ

home/utils.py
megh-k: Oh yes, so I actually did not mean to hardcode this but I also don't know enough about wagtail to figure it out so I meant to ask Codie but it slipped my mind :) thank you! @codiebeulaine could you advise?

home/utils.py
megh-k: the reason I decided to go with the x's and set them later was because SXD wants these long strings with lots of points and numbers that depend on the parent as well as other children and at the moment, the row has no awareness of what came before, so I thought it might be easier (albeit way uglier) to iterate over the "sheet" afterwards and set the numbering strings. I am really not sure though, I did consider adding a value that gets passed into the `add_children` function but I am really not sure, I think it might be an option to look into again when/if the dataclass gets implemented?

home/utils.py
rudigiesler: Ah I see, so it doesn't actually act like a dictionary. Yeah happy to leave it like this, maybe add a comment as to why we're doing it like this.

home/utils.py
rudigiesler: Yeah, definitely happy for it not to happen now. Maybe the dataclass can include it's "position" (parent, depth, etc), and then when iterating through the rows we can easily know what to put here.

codiebeulaine (approved):

PR #77, opened 2022-08-24
Add dataclass for the contentsheet rows

megh-k:


home/utils.py
```diff
@@ -445,163 +349,60 @@ def style_sheet(wb: Workbook, sheet: Worksheet) -> Tuple[Workbook, Worksheet]:
...
-    temp_sheet: List[list], children: PageQuerySet, queryset: PageQuerySet
+    temp_sheet: List[list],
+    children: PageQuerySet,
+    queryset: PageQuerySet,
+    parent_structure: str,
 ) -> List[list]:
     """Recursive function that traverses the children of a page with a depth first search algorithm"""
-    for child in children:
+    for index, child in enumerate(children, 1):
         content_page = ContentPage.objects.filter(id=child.id).first()
```
codiebeulaine, line 372: if this doesn't exist will there be an error?

home/utils.py
```diff
@@ -631,3 +432,284 @@ def export_csv_content(queryset: PageQuerySet, response: HttpResponse) -> None:
...
+    media_link: str = ""
+    related_pages: str = ""
+    tuple_of_extra_rows: tuple = ()
+
+    def __init__(
+        self,
+        page: ContentPage,
+        structure_string: str,
+        side_panel_message_number: int = 0,
+    ) -> None:
```
rudigiesler, line 472: I think with dataclasses, the convention is to let the dataclass create the `__init__` function. And if you need to create an instance of it some other way, then you use `classmethod`s, eg.
```
@dataclass
class ContentSheetRow:
    side_panel_column_1: str = ""
    ...
    
    @classmethod
    def from_page(cls, page: ContentPage, structure_string: str, ...) -> "ContentSheetRow":
        return cls(
            side_panel_column_1=...,
            ...
        )
```

home/utils.py
```diff
@@ -631,3 +432,284 @@ def export_csv_content(queryset: PageQuerySet, response: HttpResponse) -> None:
...
+                    if "next_prompt" in msg.value
+                    else None,
+                }
+            )
+        return msgs
+
+    def __set_structure(self, structure_string: str):
+        """Sets the sidebar Menu/Sub level based on structure string length,
+        for example, Sub 1.2.1 is a level 3 message"""
+        if "menu" in structure_string.lower():
```
codiebeulaine, line 565: will all index page shave menu in it? can we not find the depth from something like if it is a child of a homepage, then level is one, and so on

home/utils.py
```diff
@@ -631,3 +432,284 @@ def export_csv_content(queryset: PageQuerySet, response: HttpResponse) -> None:
...
+            self.triggers,
+            self.locale,
+            self.next_prompt,
+            self.image_link,
+            self.doc_link,
+            self.media_link,
+            self.related_pages,
+        ]
+
+    def __format_list_from_query_set(self, unformatted_query: PageQuerySet) -> str:
```
rudigiesler, line 534: Is there a reason we're using these super private double underscore methods?

home/utils.py
megh-k: `add_children` gets called in 2 places, both where `if content_page.has_children` is used so there will be a list of children only if there are children, so, I dont think it will ever not exist? but I am not sure

home/utils.py
megh-k: the `structure_string` I passed through at the point `for index, page in enumerate(main_menu_pages, 1):` contains the string menu, the assumption I made from the template sheet I was sent that they only want "Menu" for the children of the homepage? but I might have misinterpreted the template

rudigiesler (commented):
Looks good! Just two small comments/questions.

The `messaging_platform_messages` feels a bit messy to me, but I don't have any ideas on how to make it neater at the moment, so happy with it as is.

home/utils.py
megh-k: Its just habit from previous jobs where the convention was to make anything that doesnt need to be accessed from outside the class a private method, would it be better to leave them public?

home/utils.py
megh-k: ah thanks, I will look into it

megh-k:
@rudigiesler `messaging_platform_messages` is super messy and I have been trying to think of ways to get rid of it, I think I have one but it doesnt seem to be any less messy. The problem is that the page has an attribute for each platform's content and the way the template it set out, SXD wants each message to be on a separate line with nothing else in the columns, which makes it awkard to check which platform has how many messages and ensuring that each message is written to its own row and numbered. I am going to keep thinking about it though and if I figure out a better way to handle it, I will come back to it :)

home/utils.py
rudigiesler: Yeah private with a single underscore is fine, the double underscore is python's super private thing, which we don't usually use much. The docs go into it a bit here: https://docs.python.org/3/tutorial/classes.html#private-variables

rudigiesler:
@megh-k Yeah the only cleaner way I could come up with was more classes. So you'll have a 
```
class Message:
    body: str
    image: str
    ...
    
    @classmethod
    def from_platform_body(cls, platform_body: ...) -> List[Message]:
        # this function is basically __get_messages
        ...
```
And a 
```
class MessagesContainer:
    whatsapp: List[Message]
    viber: List[Message]
    ...
    
    def find_first_attachment(self, index: int, attachment_type: str) -> str:
        for message_list in [self.whatsapp, self.viber, ...]:
            try:
                value = getattr(message_list[index], attachment_type)
                if value:
                    return value
            except IndexError:
                continue
        return ""
```
Just to split those functions out into some logic classses.

Not sure if that helps? Or if you have any better ideas of organising things...

codiebeulaine (approved):

PR #82, opened 2022-09-07
Upgrade to Python 3.10, Wagtail 4.0.1, Django 4.1

megh-k:
**Upgrade**
Python 3.8 -> 3.10
Wagtail 3.0.1 -> 4.0.1
Django 4.0 -> 4.1


codiebeulaine (approved):

home/panels.py
```diff
@@ -4,6 +4,14 @@
...
     class BoundPanel(Panel.BoundPanel):
+        template_name = "panels/page_rating_panel.html"
+
+        def get_context_data(self, parent_context=None):
+            context = super().get_context_data(parent_context)
+            context["page_value"] = self.instance.page_rating
+            context["revision_value"] = self.instance.latest_revision_rating
+            return context
+
         def render(self):
```
rudigiesler, line 15: Do we need to change or remove this function now? So that it uses the template instead? Otherwise this is template is duplicated in this function.

home/templates/reports/page_view_report.html
```diff
@@ -1,108 +1,43 @@
...
-                            <div class="dropdown-toggle">{% icon name="arrow-down" %}</div>
-                            <ul>
-                                <li><a  class="button bicolor button--icon" href="{{ view.csv_export_url }}">{% icon name="download" wrapped=1 %}{% trans 'Download CSV' %}</a></li>
-                            </ul>
-                        </div>
-                    {% endif %}
-                {% endblock %}
-            </div>
-        </div>
-    </header>
```
rudigiesler, line 32: I'm not sure why we're removing this header, I don't see it being put anywhere else. But I don't know these templates well enough to know how everything links together, just wanted to raise it in case we're losing something here.

rudigiesler (commented):
Looks good! Just two small questions of things I'm not sure of

home/templates/reports/page_view_report.html
codiebeulaine: the header is defined in Wagtail's other templates, we are inheriting from those templates and only overriding block listing. They have done a major UI overhaul in this release

home/panels.py
megh-k: @rudigiesler I think you are right, I went and looked, `render_as_object` is deprecated and the render shouldn't be necessary if there is a template, I have removed both :)

codiebeulaine (approved):

PR #83, opened 2022-09-11
Update import

codiebeulaine:
## Purpose
_Update content import._

#### Steps
- [x] Update to new csv format
- [x] Add XLSX import
- [x] Add Triggers
- [x] Add Quick Replies
- [x] Update UI 
- [ ] Add error/success messages


home/utils.py
```diff
@@ -46,53 +56,30 @@
...
-                        msg_blocks.append(("next_prompt", blocks.TextBlock()))
-
-                if row:
-                    block = blocks.StructBlock(msg_blocks)
-                    block_value = block.to_python(data)
-                    struct_blocks.append((type_of_message, block_value))
-            return struct_blocks
-        else:
+        for raw in messages:
+            if "image_title" in raw and raw["image_title"]:
```
megh-k, line 72: SXD asked if we could use image_link so I made the export work with the link instead of the title, so there wont be an image title. We will also need the document and media

home/utils.py
```diff
@@ -204,37 +181,101 @@ def add_viber(row, page=None):
...
+            if next_row["messenger_body"] not in ["", None]:
+                messenger_messages.append(next_row["messenger_body"])
+            if next_row["viber_body"] not in ["", None]:
+                viber_messages.append(next_row["viber_body"])
+            counter += 1
 
+        contentpage = add_web(row)
+        contentpage = add_whatsapp(row, whatsapp_messages, contentpage)
+        contentpage = add_messenger(row, messenger_messages, contentpage)
+        contentpage = add_viber(row, viber_messages, contentpage)
```
megh-k, line 274: We also need the related pages

codiebeulaine: related pages in separate PR

home/management/commands/import_csv_content.py
```diff
@@ -1,31 +0,0 @@
-from django.core.management.base import BaseCommand
```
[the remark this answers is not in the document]
codiebeulaine: do we need it? If it's running in the UI?

home/views.py
```diff
@@ -117,36 +119,44 @@ def get_context_data(self, **kwargs):
...
-        )
+        import_content(self.file, self.file_type, self.purge, self.locale)
+
+
+class UploadView(View):
+    form_class = UploadFileForm
+    template_name = "upload.html"
 
+    def get(self, request, *args, **kwargs):
+        loading = "ContentUploadThread" in [th.name for th in threading.enumerate()]
```
rudigiesler, line 138: The issue with using threads like this, is that we're assuming we'll only ever have a single process running. If we have multiple processes, then we're not guaranteed that the http request will always go to the same one.

I think we can leave it for now, but for future we might want to think about storing this state in the database instead.

home/utils.py
```diff
@@ -204,37 +181,101 @@ def add_viber(row, page=None):
...
-
-    reader = csv.DictReader(io.StringIO(csv_file))
-    for row in reader:
+    file = file.read()
+    lines = []
+    if filetype == "XLSX":
+        wb = load_workbook(filename=BytesIO(file))
+        ws = wb.active
+        ws.delete_cols(1, 7)
+        ws.delete_rows(1, 2)
```
rudigiesler, line 226: I'm wondering if we can name these columns and rows? I'm not sure what they represent, but something like `NOTES_COLUMN = 7`, `TRANSLATOR_COMMENTS_ROW = 2`, and then use those variables here instead

home/management/commands/import_csv_content.py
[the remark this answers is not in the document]
erikh360: I guess not if we're 100% sure the new way works

home/utils.py
```diff
@@ -204,37 +181,101 @@ def add_viber(row, page=None):
...
-        contentpage = add_messenger(row, contentpage)
-        contentpage = add_viber(row, contentpage)
+        whatsapp_messages = [row["whatsapp_body"]]
+        messenger_messages = [row["messenger_body"]]
+        viber_messages = [row["viber_body"]]
+        counter = 1
+        while True:
+            try:
+                next_row = lines[index + counter]
+            except Exception:
```
rudigiesler, line 259: We probably want to restrict this to `IndexError`

home/utils.py
```diff
@@ -204,37 +181,101 @@ def add_viber(row, page=None):
...
         row = clean_row(row)
-        contentpage = add_web(row)
-        contentpage = add_whatsapp(row, contentpage)
-        contentpage = add_messenger(row, contentpage)
-        contentpage = add_viber(row, contentpage)
+        whatsapp_messages = [row["whatsapp_body"]]
+        messenger_messages = [row["messenger_body"]]
+        viber_messages = [row["viber_body"]]
+        counter = 1
+        while True:
```
rudigiesler, line 256: Maybe this could be `for next_row in lines[index + 1:]`, then we can remove the try/except

home/views.py
codiebeulaine: but they won't be able to access the import page through the UI if while there is a process running it disables the form?

home/utils.py
codiebeulaine: Meg is actually working on making this dynamic in which case I don't think we will be able to name them. These rows represent the level the item is on. and with dynamic levels coming soon, we will have to remove an arbitrary amount of columns

home/views.py
rudigiesler: Say you've scaled your webserver up, say to two processes, A and B. You make a web request to start the import, it goes to A, which starts a thread on A. Then the browser makes another request to check the status of the import, this then gets send to B. B doesn't have any threads running, so `loading` is `False`, so the page redirects, even though the import task is still running on A.

By default gunicorn has a single sync worker, so this will work. But if you need to support a higher number of requests, you might want to increase the amount of workers, or even increase the number of instances (docker containers). Since there's no communication between these, they might even be on different machines, you'll want a central place to store this state. Database is good because we don't need to add another dependancy, but this could also be a shared cache like Redis.

codiebeulaine: ah I see. Adding that as an Issue on GitHub so we don't forget

home/utils.py
```diff
@@ -2,25 +2,36 @@
...
+from wagtail.images.blocks import ImageChooserBlock
+from home.models import (  # isort:skip
+    ContentPage,
+    ContentPageIndex,
+    ContentQuickReply,
+    ContentTrigger,
+    HomePage,
+    MediaBlock,
+    Page,
+)
```
rudigiesler, line 33: Have a look at https://github.com/praekeltfoundation/vumi2/blob/main/pyproject.toml#L48-L50 if you want isort and black to work together

rudigiesler (approved):

megh-k (approved):

PR #89, opened 2022-09-22
Dont save whatsapp messages if validation fails 

megh-k:
Adding a validation to contentpage to ensure whatsapp message with attachment is shorter than 1024 chars

A `clean` method on the `ContentPage` object validates the whatsapp block to ensure that a message with an image, document or media cannot exceed 1024 characters (tested and python counts emoji's as 1 character)

An error banner is raised at the top with a list of errors - I was unable to get the errors onto a block level

whatsapp message validation failed

### DIFF
--- a/home/models.py
+++ b/home/models.py
@@ -40,6 +40,10 @@ class WhatsappBlock(blocks.StructBlock):
         max_length=20, help_text="prompt text for next message", required=False
     )
 
+    class Meta:
+        icon = "user"
+        form_classname = "whatsapp-message-block struct-block"
+
 
 class ViberBlock(blocks.StructBlock):
     image = ImageChooserBlock(required=False)
--- a/home/models.py
+++ b/home/models.py
@@ -1,4 +1,5 @@
 from django.conf import settings
+from django.core.exceptions import ValidationError
 from django.db import models
 from modelcluster.contrib.taggit import ClusterTaggableManager
 from modelcluster.fields import ParentalKey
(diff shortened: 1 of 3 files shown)


home/models.py
```diff
@@ -340,6 +341,29 @@ def whatsapp_template_name(self):
...
 
+    def clean(self):
+        message_with_media_length = 1024
+        errors = []
+        for message in self.whatsapp_body:
+            if (
+                (
+                    "image" in message.value
+                    and "document" in message.value
+                    and "media" in message.value
```
rudigiesler, line 352: Will all these fields be in `value`, even if there's only a `document`, will the `image` and `media` fields also be present?

home/models.py
```diff
@@ -340,6 +341,29 @@ def whatsapp_template_name(self):
...
+                    or message.value["document"]
+                    or message.value["media"]
+                )
+                and len(message.value["message"]) > message_with_media_length
+            ):
+                errors.append(
+                    f"A WhatsApp message with media cannot be longer than {message_with_media_length} characters long, your message is {len(message.value['message'])} characters long"
+                )
+        if errors:
+            raise ValidationError(errors)
```
rudigiesler, line 365: Should we maybe have a unit test, to ensure that the error gets raised when we expect it to?

home/models.py
megh-k: When the import PR is merged in, I would like to take this out, at the moment the import doesnt create objects under those 3 keywords, it only creates for `"image"` but if `"image"` is present in the block and the other two arent, the clean fails. This is just a work-around for the tests and using the import. But pages created by hand, and pages created with the updated import will have all three objects in the block

home/models.py
megh-k: yes we should probably have unit tests for everything :D

codiebeulaine (approved):

PR #94, opened 2022-11-03
hacky fix for encoding issues in the import due to xlsx format

megh-k:
It appears that excel is replacing the line separator used by unix based systems (`\n`) with the  line separator used by windows based systems (`\r\n`) but it decodes as `_x000D_` plus a line separator, a similar issue can be found [here](https://stackoverflow.com/questions/71574319/how-to-remove-newline-characters-so-it-doesnt-produce-x000d-when-reading-exce). 

I think this needs a better solution but I have been playing around with the file for a bit but I can't seem to get anything better than this

### DIFF
--- a/home/utils.py
+++ b/home/utils.py
@@ -257,7 +257,9 @@ def import_content(file, filetype, purge=True, locale="en"):
         for row in ws.iter_rows():
             row_dict = {}
             for index, x in enumerate(row):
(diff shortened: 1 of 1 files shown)


rudigiesler (approved):
Looks good to me. I don't think there's much else we can do here, and `_x000D_` will probably not match anything valid in cells. Also, removing them is what we want, because I think for whatsapp content we want `\n` for new lines, not `\r\n`

PR #95, opened 2022-12-01
Content variations on whatsapp

megh-k:


home/models.py
```diff
@@ -24,18 +25,51 @@
...
 class MediaBlock(AbstractMediaChooserBlock):
     def render_basic(self, value, context=None):
         pass
 
 
+def get_choices():
+    if SiteSettings.objects.first():
+        return [
+            (choice, choice)
+            for choice in SiteSettings.objects.first().content_variations_options
```
codiebeulaine, line 48: We shouldn't have SSiteSettings.objects.first here, we should get the site settings for the correct site

KaitCrawford: Follow-up PR has changed this to use the settings for the default site as discussed

home/models.py
```diff
@@ -24,18 +25,51 @@
...
+        max_lenth=4096, help_text="each message cannot exceed 4096 characters."
+    )
+
+
 class WhatsappBlock(blocks.StructBlock):
     image = ImageChooserBlock(required=False)
     document = DocumentChooserBlock(icon="document", required=False)
     media = MediaBlock(icon="media", required=False)
     message = blocks.TextBlock(
         max_lenth=4096, help_text="each message cannot exceed 4096 characters."
```
codiebeulaine, line 70: something to note for future is with the new changes, we should add check so that if message has media, the char limit is lower

home/models.py
```diff
@@ -24,18 +25,51 @@
...
+        ]
+    return []
+
+
+class VariationBlock(blocks.StructBlock):
+    variation_message = blocks.ChoiceBlock(
+        required=False,
+        choices=get_choices,
+        default=None,
+        help_text="Add variation fields to the site settings if you'd like to see them here",
```
codiebeulaine, line 58: if we can, we could add this help text into a _() translation thing but happy to do that in maintenance

codiebeulaine (approved):

PR #96, opened 2022-12-05
Use StreamField to manage pre-defined profile values in site settings

KaitCrawford:
## Purpose
Builds on previous work but allows multiple profile fields to be provided with a set list of valid values.


home/models.py
```diff
@@ -41,24 +62,49 @@ def render_basic(self, value, context=None):
...
 
-def get_choices():
+def get_valid_profile_values(field):
     if SiteSettings.objects.first():
-        return [
-            (choice, choice)
-            for choice in SiteSettings.objects.first().content_variations_options
-        ]
+        profile_values = {}
+        for profile_block in SiteSettings.objects.first().profile_field_options:
```
codiebeulaine, line 68: we shouldn't get first site settings objects but correct settings for site. I think you can do SiteSettings.for_site(site_object) iirc or you can get site for request

home/models.py
```diff
@@ -41,24 +62,49 @@ def render_basic(self, value, context=None):
...
+            profile_values[profile_block.block_type] = [b for b in profile_block.value]
+        return profile_values[field]
     return []
 
 
+def get_gender_choices():
+    # Wrapper for get_profile_field_choices that can be passed as a callable
+    choices = {k: v for k, v in GENDER_CHOICES}
```
codiebeulaine, line 76: what is k and v?

home/constants.py
```diff
@@ -0,0 +1,21 @@
...
+]
+
+AGE_CHOICES = [
+    ("15-18", "15-18"),
+    ("19-24", "19-24"),
+    ("empty", "Empty"),
+]
+
+RELATIONSHIP_STATUS_CHOICES = [
+    ("in_a_relationship", "In a Relationship"),
```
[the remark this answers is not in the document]
KaitCrawford: I've just put down the ones that we're using for YAL. I don't think I should be trying to figure out all the combinations we might need in the future. But it does raise the question of how we'll handle multiple options that mean the same thing to a project. I'll put that down for post-launch refinements though.

home/models.py
KaitCrawford: They are variables for the key and value. They are being used for the list comprehension so they don't exist outside of this line

home/models.py
```diff
@@ -41,24 +68,54 @@ def render_basic(self, value, context=None):
...
 
 
-def get_choices():
-    if SiteSettings.objects.first():
-        return [
-            (choice, choice)
-            for choice in SiteSettings.objects.first().content_variations_options
-        ]
+def get_valid_profile_values(field):
+    sitesettings = Site.objects.filter(is_default_site=True).first().sitesettings
```
codiebeulaine, line 72: get_object_or_404 might be better here? because if this does not exist, then this will throw an error NoneType has no sitesettings. 

or maybe this would not break if you do like

SiteSettings.for_site(Site.object.get(default_site=True)

only one site can be default site. so the get will not fail unless there is no site

home/models.py
codiebeulaine: 👍

home/models.py
KaitCrawford: coolio. I don't think it's even possible to get here if there isn't a default site set but rather safe than sorry

codiebeulaine (approved):

PR #109, opened 2023-01-10
search ordered content with profile fields

megh-k:
## Purpose
Make ordered content sets searchable on profile fields, `age`, `relationship`, `gender`

## Approach
add functions that are called by `SearchFields` on the model to make those fields searchable

- [x] Not sure this will apply to existing objects
We can run the following to update indexing `./manage.py update_index`

## Learning
Followed [this](https://stackoverflow.com/questions/72823305/wagtail-search-fields-on-snippet-with-foreign-key)


KaitCrawford (approved):
👍

PR #110, opened 2023-01-16
Update import/export to include message variations on whatsapp messages

megh-k:
Add export and import for variation messages where messages are added as a  separate row with the page ID and message number. 
variation title is generated as `profile_field_key: profile_field_value`

TODO: 
- [x] Add export
- [x] Add import

Exported sheet:
Screenshot 2023-01-23 152855

### DIFF
--- a/home/utils.py
+++ b/home/utils.py
@@ -187,7 +187,7 @@ def import_content(file, filetype, purge=True, locale="en"):
         else:
             home_page.add_child(instance=page)
 
-    def add_web(row, page, pk=None):
+    def add_web(row, page, pk):
         if "web_title" not in row.keys() or not row["web_title"]:
             return
         if page:
@@ -207,7 +207,7 @@ def import_content(file, filetype, purge=True, locale="en"):
             locale=home_page.locale,
         )
 
-    def add_whatsapp(row, whatsapp_messages, page=None, variation_messages=[], pk=None):
+    def add_whatsapp(row, whatsapp_messages, pk, page=None, variation_messages=[]):
         if "whatsapp_title" not in row.keys() or not row["whatsapp_title"]:
             return page
 
@@ -230,7 +230,7 @@ def import_content(file, filetype, purge=True, locale="en"):
             )
             return page
 
-    def add_messenger(row, messenger_messages, page=None, pk=None):
+    def add_messenger(row, messenger_messages, pk, page=None):
         if "messenger_title" not in row.keys() or not row["messenger_title"]:
             return page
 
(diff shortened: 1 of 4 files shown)


rudigiesler (commented):
One small comment, otherwise looks good!

home/utils.py
```diff
@@ -600,26 +655,63 @@ def from_platform_body_element(
...
+        profile_fields = []
+        for profile_field in variation.get("variation_restrictions").raw_data:
+            profile_fields.append(f"{profile_field['type']}: {profile_field['value']}")
+        message.title = ", ".join(profile_fields)
+        return message
+
+
+@dataclass
+class VariationMessageList:
+    variations: Tuple[VariationMessage]
```
rudigiesler, line 676: You've specified this as a tuple type, but everywhere I see it being used, it's being used with a list, so maybe it should be `List` instead?

home/utils.py
```diff
@@ -600,26 +655,63 @@ def from_platform_body_element(
...
+    @classmethod
+    def from_platform_body_element(cls, whatsapp_msg: blocks.StreamValue.StreamChild):
+        variations = []
+        for variation in whatsapp_msg.value["variation_messages"]:
+            variations.append(VariationMessage.from_variation(variation=variation))
+        return cls(variations)
+
+    def __iter__(self):
+        for value in self.__dict__.values():
+            yield value
```
rudigiesler, line 687: I'm unsure where we're using this `__iter__`? It's a bit of a strange way this is implemented, since the only value this class has is `variations`, this will just yield a single item, the value stored at variations. So why not just `yield self.variations`? Also, it might make more sense to do a `for message in self.variations: yield message`, so that we can do a `for variation in variation_message_list`, instead of a
```
for variations in variation_message_list:
  for variation in variations:
     ...
```

home/utils.py
```diff
@@ -306,20 +359,41 @@ def get_side_panel_width(ws, column_heading):
...
             if next_row["messenger_body"] not in ["", None]:
                 messenger_messages.append(next_row["messenger_body"])
             if next_row["viber_body"] not in ["", None]:
                 viber_messages.append(next_row["viber_body"])
-
-        contentpage = add_web(row)
-        contentpage = add_whatsapp(row, whatsapp_messages, contentpage)
-        contentpage = add_messenger(row, messenger_messages, contentpage)
-        contentpage = add_viber(row, viber_messages, contentpage)
+        exiting_contentpage = ContentPage.objects.filter(pk=pk).first()
```
KaitCrawford, line 375: What if the existing content page is a completely different page? Will the current content be overwritten?
Eg we decide we want YAL to have a section for teen mothers so we import content from momconnect, page with id 3  in the momconnect data will be very different to the page with id 3 in the YAL data.

megh-k: Yeah this is the case, but SXD wants to be able to set the page ID's and I don't want to check things on title's because with projects like OOY the titles changed a few times. I will speak to Helen and find out what she'd like but I think the best might be to add `page_id`'s to the contentpage view and allow the admin to sort and search on page id or something

PR #112, opened 2023-02-07
Better error logging

rudigiesler:
## Purpose
In order to be able to identify and fix issues quicker, we want better error logging.

## Approach
For better error logging, we've chosen Sentry.

Firstly, we disable debug mode in production, since this is a security and performance issue. To do that, we needed to:
- Make this project an installable package
- Use the praekeltfoundation/docker-django-bootstrap docker image as the base image. This gives us nginx to serve static files, and we don't need to manage running nginx or gunicorn in the container. It also runs our migrations for us.
- Collect static files in the docker image build process, for nginx to serve.
- Make DEBUG configurable, default to False

Then, we added Sentry. This was added in a way that if SENTRY_DSN is set, then sentry is loaded and configured, otherwise it is not.


megh-k (dismissed):

KaitCrawford (approved):
👍

PR #113, opened 2023-02-15
update import export to use slugs to update pages

megh-k:
## Purpose
import currently doesnt update pages
Import also only adds new tags, and adds translation tags to tags if no tags are present

## Approach
pages updated on slugs 🐌


home/utils.py
```diff
@@ -191,11 +193,20 @@ def add_parent(row, page, home_page):
         else:
             home_page.add_child(instance=page)
 
-    def add_web(row):
+    def add_web(row, page):
```
KaitCrawford, line 196: Shouldn't page here have a default, like the `add_whatsapp` definition has `page=None`?

home/utils.py
```diff
@@ -365,7 +380,9 @@ def get_side_panel_width(ws, column_heading):
             if next_row["viber_body"] not in ["", None]:
                 viber_messages.append(next_row["viber_body"])
 
-        contentpage = add_web(row=row)
+        exiting_contentpage = ContentPage.objects.filter(slug=slug).first()
```
KaitCrawford, line 383: I think this will throw an ObjectNotFound error or something if there isn't a page with the given slug 🤔

KaitCrawford (commented):
Some uncertainties on my part that I want to check

home/tests/test_api.py
```diff
@@ -34,7 +34,7 @@ def test_tag_filtering(self):
         # it should return all pages for no tag
         response = self.client.get("/api/v2/pages/")
         content = json.loads(response.content)
-        self.assertEquals(content["count"], 59)
+        self.assertEquals(content["count"], 5)
```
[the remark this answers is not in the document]
megh-k: @KaitCrawford yes, the test data was 1000+ lines and given how often the export/import has been changing, it is very tedious for me to manually edit the 1000+ lines each time, I do want to make a better and more rigorous content sheet that isnt just the WHO content sheet, it did also occur to me while editing it that it might not be the best to have this active content sitting in a test

home/utils.py
megh-k: @KaitCrawford I did test this using a slug that was not in the database and it did work and successfully created the page with that slug

KaitCrawford: ok

KaitCrawford (approved):
👍 when the tests are green

PR #114, opened 2023-02-15
Add related_pages to import

megh-k:
## Purpose
- Import related pages
- translation tags were added back to tags

## Approach
- Using the related_pages slugs in the content sheet, the related pages block was imported
- tags are cleared before we add new tags to stop tags accumulating and allowing us to update tags via the content sheet 

Screenshot 2023-02-15 161256

### DIFF
--- a/home/utils.py
+++ b/home/utils.py
@@ -3,6 +3,7 @@ import csv
 import io
 from dataclasses import dataclass
 from io import BytesIO
+from json import dumps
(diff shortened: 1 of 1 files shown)


KaitCrawford (approved):
👍 looks good 🍰

KaitCrawford:
Are there no tests for the imports?

PR #115, opened 2023-02-16
Feature/yal 726 import export of ordered content sets

KaitCrawford:
## Purpose
We need to be able to export and import ordered content sets.

## Approach
This adds util methods to export and import ordered content sets.

- [x] Still need to do the UI

## Learning
Largely copied the import and export framework used for the content pages


home/utils.py
```diff
@@ -623,6 +625,104 @@ def export_csv_content(queryset: PageQuerySet, response: HttpResponse) -> None:
         writer.writerow(row)
 
 
+def get_serialized_ordered_set(
```
megh-k, line 628: I think we should consider pulling the imports and exports for each thing into a different place? the utils file is become really large and cumbersome

rudigiesler (dismissed):

codiebeulaine (approved):

PR #116, opened 2023-02-20
Add management command for migrating related tags

rudigiesler:
## Purpose
Related tags are considered depricated, and we want to use related pages instead.

## Approach
This adds a management command to set the related pages from the related tags, for any content that is not yet using the related pages.


KaitCrawford (approved):
👍 🍰

PR #121, opened 2023-02-28
Handle invalid IDs on related tags

rudigiesler:
## Purpose
Previously, the management command trusted that the IDs in the related tags existed, but this is not the case if there are some broken tags.

This leaves the related pages in a broken state, so we rather make sure that the ID in the tag exists, before adding it to the related pages.

## Approach
Before setting the related pages, we filter the IDs by whether a page in the database exists for that ID.

### DIFF
--- a/home/management/commands/change_related_tag_to_related_page.py
+++ b/home/management/commands/change_related_tag_to_related_page.py
@@ -1,6 +1,7 @@
 import json
 
 from django.core.management.base import BaseCommand
+from wagtail.models import Page
 
 from home.models import ContentPage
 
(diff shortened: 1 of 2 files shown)


home/management/commands/change_related_tag_to_related_page.py
```diff
@@ -22,15 +23,23 @@ def add_arguments(self, parser):
...
-        pages = ContentPage.objects.live().filter(
-            tags__name__startswith=self.TAG_PREFIX
+        pages = (
+            ContentPage.objects.live()
+            .filter(tags__name__startswith=self.TAG_PREFIX)
+            .distinct()
         )
         for page in pages:
-            related_pages = set(p.value.id for p in page.related_pages)
+            related_pages = set(p.value and p.value.id for p in page.related_pages)
```
KaitCrawford, line 32: Isn't this going to be a set of boolean values?

rudigiesler: If `p.value` is `None`, then it's `None`, but if `p.value` is truthy, then it's `p.value.id`, which is an integer. Basically `p.value.id if p.value else p.value`. Not sure if an `if` is a bit clearer here?

KaitCrawford: ok, thanks. I do think an `if` would be clearer

KaitCrawford (approved):
👍

KaitCrawford (approved):
👍 thanks

PR #125, opened 2023-03-09
Don't submit duplicate whatsapp templates

rudigiesler:
## Purpose
When submitting templates, whatsapp rejects the template if the content is identical to a previous template. So for those cases, we don't want to submit templates with the same content.

## Approach
Firstly, we need to separate the template name from the revision ID, and instead store the template name in the database. This will allow us to have multiple revisions point to the same template. In order to do this, we need to:
- Add the whatsapp_template_name field to the ContentPage
- Fill that field (and also in the revisions) with what it would've been in the past (combination of title and revision ID)
- Generate a template name, and store it on the model, when submitting a new template
- Don't submit a new template if the content (template body and quick reply buttons) are the same as before. Rather use the existing template name
- Add the whatsapp template name to the import/export

## Caveats
- Currently for the import/export, we just set the whatsapp template name on the model, we don't try to submit the template. This is probably what we want most of the time, but to make it more robust we could ask the whatsapp API for the list of templates, and if a template with that name doesn't exist, then we create it during import


KaitCrawford (dismissed):
👍 this looks really good

KaitCrawford:
Sorry, could you add the template name to the API endpoint as well pls? Unless you were planning on doing that in another PR?

rudigiesler:
@KaitCrawford Added it to the API like so:
image

This is only added to the details page, and only for whatsapp, that made the most sense for me.

KaitCrawford (approved):
👍 Thanks

PR #128, opened 2023-03-29
Fix setting is_whatsapp_template should submit template

rudigiesler:
## Purpose
If a message wasn't set as a template, but then a new revision comes in that sets it as a template, it should be submitted, even though the content hasn't changed.

## Approach
When we check if the body and buttons are the same, we should also check if the value of is_whatsapp_template is the same, so that if it changes from False to True, then we also submit a template


KaitCrawford (approved):
👍

PR #133, opened 2023-06-11
Potential Word Embedding

codiebeulaine:
## Purpose
_This PR serves as a prototype on how we could implement word embedding on ContentRepo and create AAQ functionality_

## Approach
_We use a different model to AAQ but with the same accuracy, and create a util in order to match knn _

- [x] Get approach approved by Squad and Daniel
- [ ] Find a better way to load model into memory or otherwise
- [ ] Formalise sending FAQs vs general search
- [ ] Test for latency

#### Write up on our approach
- [Word Embedding Research](https://docs.google.com/document/d/11jwa2z-NWCC9iIDIPvUjZTtrSv3gPY-5aceDAqe77cQ/edit#heading=h.ls67r7pk8udw)


home/utils.py
```diff
@@ -67,6 +70,26 @@
...
 
 
+def cosine_similarity(A, B):
+    return float((np.dot(A, B) / (np.linalg.norm(A) * np.linalg.norm(B))))
+
+
+def retrieve_top_n_content_pieces(user_input, queryset, n=5, content_type=None):
+    # similar_embeddings = [{'faq_name':, 'faq_content':, 'embedding':}, ...] # We need to filter by content type and then retrieve their embeddings
+    # Generate embedding for user text
+    user_embedding = model.encode([user_input])
```
PaulEloffPraekelt, line 80: Can we update user_embedding to include text preprocessing:

user_embedding = model.encode([emoji_pattern.sub(r"", user_input).replace("\n\n", " ").replace("*", "")])

home/utils.py
```diff
@@ -67,6 +70,26 @@
...
+
+    documents_retrieved = []
+    for page in queryset:
```
PaulEloffPraekelt, line 83: In this for loop, I want to introduce a confidence threshold.

If the highest similarity_score found is less than 0.1, we need rather return a message to the user that explains that we do not have an answer for their question, and perhaps they can try to ask it differently. If the highest similarity_score found is more than 0.1 but less than 0.2, we can return the top n content pieces, but we should also add a disclaimer that we aren't very confident in our response, and again, that the user may want to try and ask the question differently.

home/api.py
```diff
@@ -42,7 +43,7 @@ def list(self, request, *args, **kwargs):
         super(ContentPagesViewSet, self).list(self, request, *args, **kwargs)
 
     def get_queryset(self):
-        queryset = super(ContentPagesViewSet, self).get_queryset()
+        queryset = ContentPage.objects.live()
```
KaitCrawford, line 46: @codiebeulaine if we're not using `get_queryset` then will we still be able to use the other built-in filters? I see that `get_queryset` calls [`get_base_queryset`](https://github.com/wagtail/wagtail/blob/main/wagtail/api/v2/views.py#L491), so if we're skipping `get_queryset` what are  losing out on?

home/utils.py
KaitCrawford: @PaulEloffPraekelt Is this comment still relevant? I remember you saying something about the newer models not needing preprocessing?

PaulEloffPraekelt: @KaitCrawford let's please exclude the preprocessing that I initially suggested to add above. Can we please have a catchup sometime to align on where you are at? Apart from the confidence threshold mentioned in my other comment, there may also be another line or two of our code that I would like to change before implementation. Also, I would like to ensure that, whichever way we are storing the data, we will be able to use it as intended downstream?

KaitCrawford: @PaulEloffPraekelt I'm currently looking to add profiling to the endpoint/codebase so that we can monitor how having the model here affects the rest of the app.
But I need to deploy this sometime this sprint as well. So let's have that conversation sooner rather than later please

home/utils.py
```diff
@@ -67,6 +70,26 @@
...
+        similarity_score = cosine_similarity(
+            user_embedding[0].tolist(), page.embedding["values"]
+        )  # Replace with your cosine similarity calculation
+        documents_retrieved.append((page.pk, page.title, page.body, similarity_score))
```
codiebeulaine, line 87: add param to search correct body

home/utils.py
KaitCrawford: We decided not to implement the confidence threshold at this time

home/utils.py
KaitCrawford: @PaulEloffPraekelt we added text preprocessing to the pages when we calculate and save their embeddings but I see now we didn't add it here on the user input. Can you confirm that we still need it here as well? (eg dealing with emojis in user questions)

rudigiesler (approved):

PR #136, opened 2023-06-21
Require authentication for all API endpoints

rudigiesler:
## Purpose
Currently all API endpoints are publically accessible. Page views allow users of the API to enter arbitrary data into the database. It is advisible to restrict access to the API to mitigate any risks that this poses.

## Approach
Requires authentication for all API endpoints.
Adds session, basic, and token authentication types
Recommended to use token authentication type


codiebeulaine (approved):

rudigiesler:
I'm going to leave this PR open so long until we're ready for v1, as to not block other work currently going on

codiebeulaine (approved):

PR #137, opened 2023-06-21
addd homepage redirect

codiebeulaine:
## Purpose
_Create redirect from homepage which is currently blank to admin page_

### DIFF
--- a/contentrepo/urls.py
+++ b/contentrepo/urls.py
@@ -3,7 +3,6 @@ from django.conf.urls.i18n import i18n_patterns
 from django.contrib import admin
 from django.urls import include, path, reverse_lazy
 from django.views.generic.base import RedirectView
-
 from rest_framework import routers
 from wagtail import urls as wagtail_urls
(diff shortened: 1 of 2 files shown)


contentrepo/urls.py
```diff
@@ -26,7 +26,7 @@
 
 urlpatterns = [
     path("django-admin/", admin.site.urls),
-    path("", include(wagtail_content_import_urls)),
```
codiebeulaine, line 29: this is going to be removed in the next PR, we don't use wagtail content import anymore

home/tests/test_views.py
```diff
@@ -15,6 +15,10 @@ class PageRatingTestCase(APITestCase):
     def setUp(self):
         self.url = reverse("contentpagerating-list")
 
+    def test_homepage_redirect(self):
+        response = self.client.get("/")
+        self.assertEquals("/admin/", response.url)
```
rudigiesler, line 20: We could also use the `assertRedirects` assertion here

rudigiesler (approved):

PR #140, opened 2023-07-05
fix-settings-issue

codiebeulaine:


home/models.py
```diff
@@ -71,16 +71,16 @@ def render_basic(self, value, context=None):
 
 def get_valid_profile_values(field):
     site = Site.objects.get(is_default_site=True)
-    if site and site.sitesettings:
-        profile_values = {}
+    site_settings, _ = SiteSettings.for_site(site)
```
rudigiesler, line 74: I think `for_site` just returns a single value.

rudigiesler (commented):
Just a small comment. Other than that looks good. 👍 when fixed.

rudigiesler (approved):

PR #144, opened 2023-07-06
Returning draft and live content on the API

megh-k:
We need to have an option to return QA content on the API, we will do this by adding the parameter `qa`

the following things are probably issues 

- The `enable_whatsapp` attribute is not saved when saving draft, so when the page that is `not_live()` is called, the `enable_whatsapp=False`, because of this, we have to add two confitional statements to the serializer to surface the whatsapp content `(page.enable_whatsapp is True or ("qa" in request.GET and request.GET["qa"] == "True"))`
- The previous way of creating the `queryset`, `queryset = super(ContentPagesViewSet, self).get_queryset()`, included `HomePage`'s and `ContentIndexPages`, as can be seen by the `test_api`, this is not the case by calling `ContentPage.objects.live()` and `ContentPage.objects.not_live()`, this might be a problem, but with both, the api doesnt find those two page types if we look for a page on the ID, so the impact of this is uncertain 
- Updated content that is saved as draft on a live page, such that the page is `live + draft` in the admin, I have not yet found a way to add that content to the API

### DIFF
--- a/home/api.py
+++ b/home/api.py
@@ -26,6 +26,7 @@ class ContentPagesViewSet(PagesAPIViewSet):
             "tag",
             "trigger",
             "page",
+            "qa",
         ]
     )
     pagination_class = PageNumberPagination
@@ -47,7 +48,12 @@ class ContentPagesViewSet(PagesAPIViewSet):
         super(ContentPagesViewSet, self).list(self, request, *args, **kwargs)
 
     def get_queryset(self):
-        queryset = super(ContentPagesViewSet, self).get_queryset()
+        qa = self.request.query_params.get("qa")
+        queryset = ContentPage.objects.live()
+
+        if qa:
+            queryset = queryset | ContentPage.objects.not_live()
+
         tag = self.request.query_params.get("tag")
         if tag is not None:
             ids = []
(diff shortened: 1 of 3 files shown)


home/api.py
```diff
@@ -47,7 +48,12 @@ def list(self, request, *args, **kwargs):
         super(ContentPagesViewSet, self).list(self, request, *args, **kwargs)
 
     def get_queryset(self):
-        queryset = super(ContentPagesViewSet, self).get_queryset()
+        qa = self.request.query_params.get("qa")
+        queryset = ContentPage.objects.live()
```
rudigiesler, line 52: If we keep the `queryset = super(ContentPagesViewSet, self).get_queryset()`, and do a `queryset = queryset | ContentPage.objects.not_live()`, does that maybe help with being able to include the home and index pages?

codiebeulaine:
who is on messenger also

home/tests/test_api.py
```diff
@@ -41,7 +41,8 @@ def test_tag_filtering(self):
         # it should return all pages for no tag
         response = self.client.get("/api/v2/pages/")
         content = json.loads(response.content)
-        self.assertEquals(content["count"], 5)
+        # exclude home pages and index pages
+        self.assertEquals(content["count"], 3)
```
codiebeulaine, line 45: can we add a test also for when QA tag is included

codiebeulaine (approved):

PR #149, opened 2023-07-11
Trying to surface published pages with draft content

megh-k:
Tags & Triggers dont created tagged objects on draft save - this is a problem when hitting the api with a tag or trigger parameter, this is not a problem in the actual content, that is why quick replies are not an issue


megh-k:
👍 🍰

home/api.py
```diff
@@ -37,15 +38,46 @@ def get(self, request, *args, **kwargs):
 
     def detail_view(self, request, pk):
         try:
-            ContentPage.objects.get(id=pk).save_page_view(request.query_params)
+            if "qa" in request.GET and request.GET["qa"] == "True":
+                instance = ContentPage.objects.get(
+                    id=pk
```
codiebeulaine, line 43: get could fail if does not exist

codiebeulaine (approved):

PR #153, opened 2023-07-13
Refactor content page import

rudigiesler:
## Purpose
I found the content page import difficult to understand quickly and difficult to write tests for, so I refactored it into a (hopefully) better layout.

## Approach
We have one class, the ContentRow, which is our representation of a row on the import document, and is agnostic as to whether it's a CSV or an XLSX. The first step is to translate the specific document type into this generic row format.

We then have another class, ShadowContentPage. This is a representation of a ContentPage model, but just using a normal class. I found it quite difficult to create and modify the wagtail models easily (eg. I couldn't figure out how to modify a single WhatsApp block in a streamfield to add a variation message, some comments online say that these elements are designed to be immutable). This class allows us to easily build up the content pages from the rows in the import, which is our second step.

The third step is to take all the ShadowContentPages, now that they're filled out, and save them to the database. This is now easier as we have all the data we need in one place, so we're just creating/replacing, not trying to update nested fields.

The final step is to link all the related pages. We unfortunately can't do this before this time, as we need the page to exist in the database before we can link to it. So this causes a second write to the database for any pages that have related pages.

Finally, all the logic has been split out into smaller, more easily digestable functions, that (hopefully) don't mix levels of abstraction.


codiebeulaine:
Good direction, we would need to update the admin import button to use this importer as well as maybe check how it affects the progress bar?

PR #165, opened 2023-08-23
Reduce docker image size

rudigiesler:
## Purpose
Reduce startup time by reducing the size of the docker image

## Approach
- Add development dependancies as optional dependancies
- Don't cache pip downloads in docker image builds
- Use CPU only pytorch (we don't run on GPUs)

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM ghcr.io/praekeltfoundation/docker-django-bootstrap-nw:py3.10-buster
 
 COPY . /app
-RUN pip install -e .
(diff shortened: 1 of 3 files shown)


rudigiesler:
Before:
image

After:
image

Saved about 2GB from the pip install, by not caching dependencies, and by installing the CPU only version of `pytorch`

The collectstatic command is still 390MB, but I don't think that's due to static files (only 13MB on my local machine), I think that's due to downloading the pytorch models on import. But on startup those get downloaded again, so we need to figure out how to either exclude them in the image, or to get it to stop redownloading them on startup.

jerith (approved):

PR #169, opened 2023-08-30
Add image to template submission

fritzbrand:
## Purpose
To allow the submission of WhatsApp templates including images

## Approach
Added a test image file that gets used in tests.  We upload this file to the database to get an image object in the CMS.  We then get the image by ID  (as we would if we got the ID from the WhatsApp message) and then pull it as part of the content


contentrepo/settings/base.py
```diff
@@ -145,6 +145,7 @@
 MEDIA_ROOT = os.path.join(BASE_DIR, "media")
 MEDIA_URL = "/media/"
 
+TEST_STATIC_ROOT = os.path.join(BASE_DIR, "home/tests/test_static")
```
rudigiesler, line 148: Do we need this? If we just do the relative path `open("home/tests/test_static/test.jpeg", ...`, does that not work just fine? Happy to keep this here if that doesn't work.

home/tests/test_whatsapp.py
```diff
@@ -53,3 +57,80 @@ def test_create_whatsapp_template_with_buttons(self):
...
+    def test_create_whatsapp_template_with_image(self):
+        img_name = "test.jpeg"
+        img_path = f"{settings.TEST_STATIC_ROOT}/{img_name}"
+
+        f = open(img_path, "rb")
+        image = Image(
+            title=img_name,
+            file=ImageFile(f, name=img_path),
+        )
+        mock_file_content = image.file
```
rudigiesler, line 71: I don't think this variable name is true anymore, or needed. (it contains the open file, not the content of that file)

home/tests/test_whatsapp.py
```diff
@@ -53,3 +57,80 @@ def test_create_whatsapp_template_with_buttons(self):
...
+                    req_files,
+                )
+            ],
+        )
+
+        template_url = "http://whatsapp/graph/v14.0/27121231234/message_templates"
+        responses.add(responses.POST, template_url, json={})
+        #  To run the WhatsAppTests on their own, the ID of 2 below should be set to 1
+        #  This is due to other content being inported if you run the full set of tests
+        create_whatsapp_template("Test-Template", "Test Body", [], 2)
```
rudigiesler, line 102: Can we not use `image.id` here, instead of the hardcoded `2`? And then get rid of the comment

home/whatsapp.py
```diff
@@ -22,16 +25,86 @@ def create_whatsapp_template(name, body, quick_replies=[]):
...
+    url = urljoin(
+        settings.WHATSAPP_API_URL,
+        "graph/v14.0/app/uploads",
+    )
+    headers = {
+        "Content-Type": "application/json",
+    }
+    images = Image.objects.all()
+    for i in images:
+        print(i.id)
```
rudigiesler, line 62: I don't think we need this `print` anymore

rudigiesler (commented):
Looks good! A few comments on things to cleanup, but the approach looks good to me

home/tests/test_whatsapp.py
fritzbrand: Fixed

home/tests/test_whatsapp.py
fritzbrand: :facepalm: Yes, tired brain wasn't thinking straight :joy:

home/whatsapp.py
fritzbrand: Yup, removed, thx

home/tests/test_whatsapp.py
rudigiesler: I'm seeing the other changes, but not this one

fritzbrand: Argh, sorry man, my earlier undo went a bit too far back 😂. It should be there now.  I tried removing the variable and doing the saved_image.file on line 82 where it is added to req_files, but it didn't like that, so I renamed it and renamed f to read_file.

rudigiesler (approved):

PR #172, opened 2023-08-31
Improved linting and static analysis

jerith:
## Purpose
Improve static analysis by switching to `ruff` and possibly adding `mypy`.


jerith:
Wagtail has no type annotations, and Django's are problematic in various ways. It may be worth working around those issues to use `mypy` anyway, but that's too much work for this PR.

rudigiesler (dismissed):

rudigiesler (approved):

PR #175, opened 2023-09-05
Add image to template submission

fritzbrand:
Missed adding the image bit to the actual submit_template call in models

### DIFF
--- a/home/models.py
+++ b/home/models.py
@@ -507,7 +507,8 @@ class ContentPage(Page, ContentImportMixin):
     def whatsapp_template_body(self):
         return self.whatsapp_body.raw_data[0]["value"]["message"]
 
-    def get_whatsapp_image(self):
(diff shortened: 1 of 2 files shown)


home/models.py
```diff
@@ -585,6 +588,7 @@ def submit_whatsapp_template(self, previous_revision):
             self.whatsapp_template_name,
             self.whatsapp_template_body,
             sorted(self.quick_reply_buttons),
+            self.get_whatsapp_image(),
```
rudigiesler, line 591: Lines 576-579, we're checking if the body, buttons, or the "is_whatsapp_template" property has changed. Maybe we should add there to check if the image has changed (or one was added)? Otherwise if you edit a message to add an image, no template will be submitted because nothing else changed.

rudigiesler (approved):

PR #176, opened 2023-09-05
Changed reading of file to (hopefully) not use absolute path in S3 bu…

fritzbrand:
When I ran the code in production (we don't have a QA yet), I got the following error. 
`Exception Type: NotImplementedError at /admin/pages/163/edit/
Exception Value: This backend doesn't support absolute paths.` 

After some reading, it would appear that for S3 buckets they suggest you use `file.name` rather than `file.path`.  Not sure if this will do the trick, but the tests pass locally


home/whatsapp.py
```diff
@@ -61,7 +61,7 @@ def get_upload_session_id(image_id):
     img_obj = get_image_model().objects.get(id=image_id)
     mime_type = mimetypes.guess_type(img_obj.file.name)[0]
     file_size = img_obj.file.size
-    file_path = img_obj.file.path
+    file_path = Path(settings.MEDIA_ROOT) / img_obj.file.name
```
rudigiesler, line 64: I'm not sure this will work either. This is still looking on a local disk for the file, not downloading it from S3. We probably want to open it directly, instead of opening a path to the file: https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html#storage

PR #182, opened 2023-09-12
Refactor content page import

jerith:
Closes #153

## Purpose
The current import/export code is difficult to understand and test, so we're restructuring it. This picks up where #153 left off, building on top of the recent testing work.

## Approach
From #153:

We have one class, the ContentRow, which is our representation of a row on the import document, and is agnostic as to whether it's a CSV or an XLSX. The first step is to translate the specific document type into this generic row format.

We then have another class, ShadowContentPage. This is a representation of a ContentPage model, but just using a normal class. I found it quite difficult to create and modify the wagtail models easily (eg. I couldn't figure out how to modify a single WhatsApp block in a streamfield to add a variation message, some comments online say that these elements are designed to be immutable). This class allows us to easily build up the content pages from the rows in the import, which is our second step.

The third step is to take all the ShadowContentPages, now that they're filled out, and save them to the database. This is now easier as we have all the data we need in one place, so we're just creating/replacing, not trying to update nested fields.

The final step is to link all the related pages. We unfortunately can't do this before this time, as we need the page to exist in the database before we can link to it. So this causes a second write to the database for any pages that have related pages.

Finally, all the logic has been split out into smaller, more easily digestable functions, that (hopefully) don't mix levels of abstraction.

### DIFF
--- a/home/tests/test_content_import_export.py
+++ b/home/tests/test_content_import_export.py
@@ -118,6 +118,8 @@ class ImportExportTestCase(TestCase):
          * We should probably set contentpage.translation_key on import instead
            of messing about with tags.
          * Do we expect imported content to have leading spaces removed?
+         * Should we set enable_web and friends based on body, title, or an
+           enable field that we'll need to add to the export?
         """
         csv_bytes = self.import_csv("home/tests/content2.csv")
         resp = self.client.get("/admin/home/contentpage/?export=csv")
--- a/home/tests/test_content_import_export.py
+++ b/home/tests/test_content_import_export.py
@@ -37,25 +37,27 @@ def remove_translation_tag_from_tags(src, dst):
     if not src["translation_tag"]:
         return src, dst
     dtags = [tag for tag in dst["tags"].split(", ") if tag != src["translation_tag"]]
-    return src, {**dst, "tags": ", ".join(dtags)}
+    return src, dst | {"tags": ", ".join(dtags)}
 
 
 @filter_both
 def ignore_certain_fields(entry):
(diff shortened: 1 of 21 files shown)


rudigiesler (dismissed):

jerith:
Line coverage of the new importer:

```
Name                           Stmts   Miss  Cover   Missing
------------------------------------------------------------
home/import_content_pages.py     326      1    99%   74
```

Line 74 handles content pages that don't have a parent, which can't happen if we our content sheet is valid.

home/tests/test_content_import_export.py
```diff
@@ -1,109 +1,487 @@
...
+            locale = str(loc)
+            url = f"{url}&locale__id__exact={loc.id}"
+        content = self.admin_client.get(url).content
+        # Hopefully we can get rid of this at some point.
+        if locale:
+            content = self._filter_export(content, locale=locale)
+        if self.format == "csv":
+            print("-v-CONTENT-v-")
+            print(content.decode())
+            print("-^-CONTENT-^-")
```
rudigiesler, line 425: Do we still need this output for the tests?

rudigiesler (commented):

home/tests/test_content_import_export.py
jerith: I added and removed it twice, so the third time I just left it in. 🤷

It's useful for figuring out whether it's the import or the export that's causing a test to fail.

rudigiesler (approved):

KaitCrawford (approved):
👍

PR #186, opened 2023-09-21
MNCH 518 whatsapp message buttons

rudigiesler:
## Purpose
We want to be able to define a list of up to 3 buttons for whatsapp messages, to be used for interactive messages

## Approach
We create a new StreamBlock for whatsapp messages, named "buttons". Here we can add up to 3 buttons of varying types.

Currently we have "next message" and "go to page" types, but we can always add more if the need arises.

We also add a data migration, to change all the `next_prompt`s to buttons, and then we removed the `next_prompt` field

image


home/tests/test_models.py
```diff
@@ -155,15 +155,13 @@ def create_message_value(
         media=None,
         message="",
         variation_messages=None,
-        next_prompt="",
```
jerith, line 158: Do we need tests for the new `buttons` field in here?

jerith (commented):
I don't see any changes to the import/export code here. Does that mean it'll still try to import `next_prompt`?

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+- Buttons StreamBlock for WhatsApp messages
+
+### Changed
+- Migrated next_prompt on WhatsApp messages to buttons
+- Removed next_prompt on WhatsApp messages
```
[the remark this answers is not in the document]
jerith: I've just checked. The exporter handles `next_prompt`, but the (current) importer apparently ignores it. Merging this branch into the import/export refactor PR gives me test failures for:
* `next_prompt` not being preserved across import->export with new importer (kind of expected because it's been removed from the model, but I'd expect the export to fail because of the missing field -- apparently we explicitly ignore those errors, which we probably shouldn't)
* (empty) `buttons` field not being present in export->import with current importer (expected because the current importer builds the content blocks generically instead of using the appropriate block models)

pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "contentrepo"
-version = "1.1.0-dev.3"
+version = "2.0.0-dev.0"
```
jerith, line 7: Do we really need to bump the major version for this?

rudigiesler: Any existing services using the API will be looking at `next_prompt` in the API, they need to change to look in the `buttons` section instead, in a completely different format (there may be multiple buttons, they might not all be next buttons). So I would say for the API users, it is a major change.

The other thing we could do is version the API. So do something like the existing API returns the first next message button as `next_prompt`, and then we have a new version that returns the full `buttons` detail. I'm not sure if that's worth the effort though.

CHANGELOG.md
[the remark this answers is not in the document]
rudigiesler: Yeah so import/export will need to be changed, but that's a bit difficult while it's still being worked on. I don't really want to make the changes to the old import/export, if it's going to be replaced anyway. Maybe we could hold off on this PR until that work is complete, then rebase/merge, and add to the new importer?

home/tests/test_models.py
rudigiesler: I've added tests for the validations on the individual button types. I'm unsure how to add tests here for compound values, adding to the dictionary doesn't work. The only test I would want to add here is to ensure you can't add more than 3 buttons

rudigiesler: Okay I've figured out the compound values, added a test for max 3 buttons.

home/import_content_pages.py
```diff
@@ -147,39 +170,29 @@ def create_shadow_content_page_from_row(self, row: "ContentRow") -> None:
...
             page.whatsapp_title = row.whatsapp_title
             if row.is_whatsapp_template_message:
                 page.is_whatsapp_template = True
                 page.whatsapp_template_name = row.whatsapp_template_name
-            page.whatsapp_body.append(
-                ShadowWhatsappBlock(
-                    message=row.whatsapp_body,
-                    next_prompt=row.next_prompt,
-                )
-            )
```
rudigiesler, line 162: This logic was being duplicated in `add_message_to_shadow_content_page_from_row`, so I've removed it from here and used that instead.

CHANGELOG.md
```diff
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+- Buttons StreamBlock for WhatsApp messages
+
+### Changed
+- Migrated next_prompt on WhatsApp messages to buttons
```
jerith, line 14: If we're deprecating `next_prompt` instead of removing it, do we still want this migration?

home/content_import_export.py
```diff
@@ -770,8 +774,39 @@ def from_platform_body_element(
...
         )
+        message.buttons = (
+            cls.serialise_buttons(platform_body_element.value["buttons"])
+            if "buttons" in platform_body_element.value
+            else ""
+        )
         return message
 
+    @classmethod
+    def serialise_buttons(cls, buttons: blocks.StreamValue.StreamChild) -> str:
```
jerith, line 785: Out of scope for this PR, but should we perhaps copy and refactor the exporter like we did for the importer?

home/import_content_pages.py
```diff
@@ -69,7 +77,10 @@ def process_rows(self, rows: list["ContentRow"]) -> None:
     def save_pages(self) -> None:
         for i, page in enumerate(self.shadow_pages.values()):
             if page.parent:
-                parent = Page.objects.get(title=page.parent, locale=self.locale)
+                # TODO: We should need to use something unique for `parent`
+                parent = Page.objects.filter(
+                    title=page.parent, locale=self.locale
+                ).first()
```
jerith, line 83: Where did this assumption break? It's never actually been valid, but having the import fail is probably better than silently reparenting pages.

home/tests/test_content_import_export.py
```diff
@@ -319,6 +319,24 @@ def remove_next_prompt(page: DbDict) -> DbDict:
...
+def remove_buttons(page: DbDict) -> DbDict:
+    if "whatsapp_body" in page["fields"]:
+        for body in page["fields"]["whatsapp_body"]:
+            body["value"].pop("buttons", None)
+    return page
+
+
+@per_page
+def remove_button_ids(page: DbDict) -> DbDict:
+    if "whatsapp_body" in page["fields"]:
```
jerith, line 332: Would it make sense to combine this, `normalise_related_page_ids`, and `normalise_body_ids` into a more generic `normalise_block_ids` filter? (I'm happy to give that a try, probably in a separate PR.)

home/tests/test_content_import_export.py
```diff
@@ -711,13 +731,20 @@ def test_variations(self, impexp: ImportExportFixture) -> None:
...
-            WABlk("Message 1", next_prompt="Next message", variation_messages=m1vars),
+            WABlk(
+                "Message 1",
+                buttons=[{"type": "next_message", "value": {"title": "Next message"}}],
+                variation_messages=m1vars,
+            ),
             WABlk(
                 "Message 2, variable placeholders as well {{0}}",
-                next_prompt="Next message",
+                buttons=[{"type": "next_message", "value": {"title": "Next message"}}],
```
jerith, line 741: Do we still have tests for `next_prompt` now that it's not gone completely? Do we support mixing buttons and `next_prompt` in the same message? (If not, we should probably enforce that somehow.)

jerith (commented):
I think we're going to need explicitly versioned import (and maybe export?) formats. Doing that in the exported files will be a bit clunky, but we can probably determine the version based on the set of headers/keys we get. Any thoughts?

CHANGELOG.md
rudigiesler: Yes I think so, it will allow people to easily move over to using the new format, instead of content admins having to manually go through all the content and add buttons.

home/import_content_pages.py
rudigiesler: This broke when using data on my local system, which is from an export of a production system. I think it is better to fail here, so maybe I'll change it back, although that will break backwards compatibility for old exports.

home/tests/test_content_import_export.py
rudigiesler: Hmm that's true. We do support mixing them (they should be treated as just separate things, and the application has to decide which one to use, to allow a smooth changeover), so I'll update this test to check that all combinations can work

home/tests/test_content_import_export.py
rudigiesler: Yeah happy for you to give it a try in a separate PR. I don't think it's high priority though, since we've only got 2 instances of us needing to do this (I prefer to wait until we have 3), and it's in tests not production code, but I think blocks are well enough defined in wagtail that we can make something generic.

CHANGELOG.md
jerith: So existing `next_prompt`s will be migrated, but then any new ones (added or imported) will need to be manually changed over at some point before `next_prompt` goes away? Is it worth adding a management command or something to do that?

jerith (approved):

CHANGELOG.md
rudigiesler: Migrating at this point, when we know `buttons` is empty, is easy to do. Doing it later on is much more difficult, eg. what do you do if there are already 3 buttons, and will require human intervention, so I'm happy leaving that for now, and working through any specific use cases (and maybe creating a management command at that point once we know more)

PR #190, opened 2023-10-09
Test against multiple postgres versions

Hlamallama:
## Purpose
Test against multiple postgres versions

## Approach

## Learning
_Describe the research stage_

#### Blog Posts


rudigiesler (dismissed):

rudigiesler:
@Hlamallama After looking at this again, I see that there's still only one check being performed. I think that's because there are two `strategy` blocks in the file, and I think the one defined later is overwriting your one. So we probably want to combine those into one strategy block

.github/workflows/build.yml
```diff
@@ -5,9 +5,14 @@ on: [push]
 jobs:
   build:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        postgres_version: ['12', '13', '14']
+        python-version: [3.10.6]
+        redis-version: [6]
```
bearnard, line 12: where is this used?

Hlamallama: deleting this PR, it was based off an outdated branch

PR #192, opened 2023-10-12
Add example support

fritzbrand:
## Purpose
WhatsApp made a change across versions of the API, that makes supplying example values mandatory, when using variables in a template

## Approach
Add support for example values in the ContentRepo, and add the functionality to include them in the template creating process


fritzbrand:
@rudigiesler the last check is taking forever, so I am going to assume it's fine for now.

Please take note of a comment on line 556 in models.py.

Also, I messed up the migration file 0037.  I reverted back to the previous migration to make some changes, but I see there were things in there about migrating stream data and the function 'copy_next_prompt_to_next_button_for_message'

Did you add those manually?

home/migrations/0037_alter_contentpage_whatsapp_body.py
```diff
@@ -76,6 +75,24 @@ class Migration(migrations.Migration):
...
+                                                        label="Example Value"
+                                                    ),
+                                                )
+                                            ]
+                                        ),
+                                        default=[],
+                                        help_text="Please add example values for all variables used in a WhatsApp template",
+                                        label="Variable Example Values",
+                                    ),
+                                ),
```
rudigiesler, line 95: For this migration, we probably don't want to edit this existing migration, because systems that already have this migration applied won't then do this additional step. What I would do here is revert the changes to this file, and run `makemigrations` again, to generate this as an `0038` migration.

home/models.py
```diff
@@ -193,6 +199,12 @@ class WhatsappBlock(blocks.StructBlock):
         "media cannot exceed 1024 characters.",
         validators=(MaxLengthValidator(4096),),
     )
+    example_values = blocks.ListBlock(
+        ExampleValuesBlock(),
```
rudigiesler, line 203: Do we need the children here to be `StructBlock`s? Could we not just do a `blocks.ListBlock(blocks.CharBlock(....))`?

home/models.py
```diff
@@ -212,19 +224,25 @@ class Meta:
 
     def clean(self, value):
         result = super().clean(value)
+        num_vars_in_msg = result["message"].count("{{")
```
rudigiesler, line 227: We could maybe use regex here to do a bit of a better match. ie. `len(re.findall("{{\d+}}", result["message"]))`

home/models.py
```diff
@@ -212,19 +224,25 @@ class Meta:
...
+        errors = {}
+        if num_vars_in_msg > 0:
+            num_example_values = len(result["example_values"])
+            if num_vars_in_msg != num_example_values:
+                errors["example_values"] = ValidationError(
+                    "The number of example values provided does not match the number of variables used in the template"
```
rudigiesler, line 233: Maybe we could add the number of example values provided, and the number of variables that we counted in the message, to this error message? eg. `The number of example values provided (2) does not match the number of variables used in the template (3)`

home/models.py
```diff
@@ -532,6 +551,20 @@ def whatsapp_template_body(self):
     def whatsapp_template_image(self):
         return self.whatsapp_body.raw_data[0]["value"]["image"]
 
+    @property
+    def whatsapp_template_example_values(self):
+        # TO CHECK IN REVIEW - I know there must be a cleaner way of doing this, but I can't seem to come up with it
```
rudigiesler, line 556: Maybe a:
```
example_values = self.whatsapp_body.raw_data[0]["value"].get("example_values", [])
return [v["value"]["example_values"] for v in example_values]
````

File: home/tests/test_content_import_export.py [hunk 6]
```diff
@@ -327,6 +327,14 @@ def remove_buttons(page: DbDict) -> DbDict:
     return page
 
 
+@per_page
+def remove_example_values(page: DbDict) -> DbDict:
+    if "whatsapp_body" in page["fields"]:
+        for body in page["fields"]["whatsapp_body"]:
+            body["value"].pop("example_values", None)
+    return page
```
rudigiesler, line 335: We probably want to add the example values to the export and import, but happy for that to happen in another PR

home/models.py
fritzbrand: 🤦 Thanks, fixed

home/migrations/0037_alter_contentpage_whatsapp_body.py
fritzbrand: Cool thanks

home/models.py
fritzbrand: Cool, thanks

home/models.py
fritzbrand: Cool, done

home/models.py
fritzbrand: Rad, thanks.  I modified it slightly as the structure changed when I removed the StructBlocks

home/tests/test_content_import_export.py
fritzbrand: Cool, will do

home/migrations/0037_alter_contentpage_whatsapp_body.py
rudigiesler: I think we still need to remove these changes here

home/whatsapp.py
```diff
@@ -40,6 +54,7 @@ def create_whatsapp_template(name, body, quick_replies=(), image_id=None):
         "language": "en_US",
         "components": components,
     }
+    print(json.dumps(data, indent=4))
```
rudigiesler, line 57: We can probably remove this print?

rudigiesler (commented):
Looks good! Just some small things, looks like the changes to the existing migration are still included in the PR

home/migrations/0037_alter_contentpage_whatsapp_body.py
fritzbrand: Sorry, I discarded the changes in git, but looks like that didn't work.  Tried copying down the contents of the file that was in main now

home/whatsapp.py
fritzbrand: Yup, sorry it snuck back in when I removed the StructBlocks and had to check what that looked like in the request

rudigiesler (approved):
Looks good!

PR #194, opened 2023-10-16
Configurable template category

rudigiesler:
## Purpose
The template category is currently hardcoded to "UTILITY". We instead want to be able to set this on a per-template basis, as its value determines limitations on sending templates, as well as the template approval process

## Approach
We add an additional field in the contentpage, under the whatsapp section in the admin interface, for setting the category. It has a default value of UTILITY.

We then use this value when submitting the template, so that it's submitted using the correct category.

We also include this value in imports/exports.


home/models.py
```diff
@@ -601,6 +613,8 @@ def submit_whatsapp_template(self, previous_revision):
                 and self.is_whatsapp_template == previous_revision.is_whatsapp_template
                 and self.whatsapp_template_image
                 == previous_revision.whatsapp_template_image
+                and self.whatsapp_template_category
+                == previous_revision.whatsapp_template_category
             ):
```
jerith, line 618: Could we factor this out into its own method?

home/content_import_export.py
```diff
@@ -238,6 +239,7 @@ def add_whatsapp(row, whatsapp_messages, page=None, variation_messages=()):
             return page
 
         whatsapp_template_name = row.get("whatsapp_template_name", "")
+        whatsapp_template_category = row.get("whatsapp_template_category") or "UTILITY"
```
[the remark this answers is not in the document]
rudigiesler: Because there could be a `whatsapp_template_category` column, but it has no value for that row. In that case if we do `row.get("whatsapp_template_category", "UTILITY")`, we'll get `""` returned, which is not a value value for this field. We want the default here to be `UTILITY`, so if they leave it blank we should choose a correct default.

home/models.py
rudigiesler: Yup, that's a good idea. I separated it out into a method that returns just the fields, for an easy equality check.

rudigiesler:
@jerith Ready for review again

jerith (approved):

jerith (approved):

PR #196, opened 2023-10-18
Better translation imports

jerith:
## Purpose
Currently, importing content that contains multiple languages is something of a disaster. The import requires a specific locale to be chosen, and all pages are imported as if they were in that locale.

Firstly, the locale-specific import should ignore any pages not in the specified locale.

Secondly, we should be able to import all pages at once and have each end up in the correct locale-specific tree.

## Approach
This PR only changes the backend import code; the UI still requires a locale to be selected.

The new import code (currently in use) now looks up the appropriate parent for each imported page based on the page's locale. It also now accepts `None` for the import locale (the new default), and treats the import locale as a filter rather then using it in place of each page's locale.

The UI hasn't been updated yet, so the import form still requires a locale to be chosen. That needs to be fixed in a future PR.

### DIFF
--- a/home/content_import_export.py
+++ b/home/content_import_export.py
@@ -74,7 +74,7 @@ logger = getLogger(__name__)
 
 
 @transaction.atomic
-def import_content(file, filetype, progress_queue, purge=True, locale="en"):
+def import_content(file, filetype, progress_queue, purge=True, locale=None):
     from .import_content_pages import ContentImporter
 
     importer = ContentImporter(file.read(), filetype, progress_queue, purge, locale)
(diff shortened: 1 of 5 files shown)


rudigiesler (commented):
Looks good!

@fritzbrand Are you happy to add the frontend changes directly to this PR, or do you want to merge this and do it in a separate PR?

jerith:
Given that there are still open questions about what we want on the frontend, I'd rather merge this (if we're happy with it) and open a new PR for the frontend work.

rudigiesler (approved):

PR #203, opened 2023-10-23
Add example values import export

fritzbrand:


rudigiesler (approved):
Looks good!

The only suggestion I have would be to treat example values similar to how we treat quick replies, triggers, and tags. That is have them be comma separated, eg. `example1,example2`, rather than json serialised, eg. `["example1","example2"]`, as it's easier to edit in a spreadsheet by content admins.

jerith:
> The only suggestion I have would be to treat example values similar to how we treat quick replies, triggers, and tags. That is have them be comma separated, eg. `example1,example2`, rather than json serialised, eg. `["example1","example2"]`, as it's easier to edit in a spreadsheet by content admins.

The problem with that is that commas are legal in example values. (Are they legal in the others as well?)

rudigiesler:
> > The only suggestion I have would be to treat example values similar to how we treat quick replies, triggers, and tags. That is have them be comma separated, eg. `example1,example2`, rather than json serialised, eg. `["example1","example2"]`, as it's easier to edit in a spreadsheet by content admins.
> 
> The problem with that is that commas are legal in example values. (Are they legal in the others as well?)

Ah right. Yes they're not legal for quick replies, triggers, tags, or related pages. So we'll have issues if example values have commas in them, so we'll probably want to ensure we're quoting and unquoting it correctly in export/import.

Maybe we can use the CSV library for our serialisation/deserialisation? That way if there are no commas in the values, we get the same format as quick replies, triggers, etc. But if there are, then it's escaped correctly?

jerith:
> Maybe we can use the CSV library for our serialisation/deserialisation? That way if there are no commas in the values, we get the same format as quick replies, triggers, etc. But if there are, then it's escaped correctly?

We already do that, which is how we have comma-separated tags (etc.) in a single field. We'd need another layer of escaping within each example for this. 🤷

fritzbrand:
Shall I merge it as it is now, and we tackle the change (if required) separately?

rudigiesler:
> Shall I merge it as it is now, and we tackle the change (if required) separately?

Yeah go ahead. It is a risk because to change it later will break things, so we could end up stuck with it, but if we can make the change before another release it should be okay.

rudigiesler:
> We already do that, which is how we have comma-separated tags (etc.) in a single field. We'd need another layer of escaping within each example for this. 🤷

Yeah I'm thinking of using CSV as the encoding for just this specific field. That means for a CSV export, this field will get CSV encoded twice, but I don't think it will be that big an issue, since it will be decoded when viewed in a spreadsheet application.

Are there any other encoding schemes you think could be a better fit than JSON?

fritzbrand:
In a call between Jeremy, Rudi and I, we decided to add validation to `example_values` to not allow commas in that field.  We will then treat it exactly like quick replies, triggers and tags wrt import/export

PR #206, opened 2023-10-25
Example values structure

fritzbrand:
Adds validation on `example_values` to prohibit commas.  Example_values is then exported as a comma separated string, rather than a list, so its easier to edit in a spreadsheet.


home/tests/test_content_import_export.py
```diff
@@ -573,19 +573,6 @@ def test_roundtrip_less_simple(self, csv_impexp: ImportExportFixture) -> None:
         src, dst = csv_impexp.csvs2dicts(csv_bytes, content)
         assert dst == src
 
-    def test_roundtrip_multiple_messages(self, csv_impexp: ImportExportFixture) -> None:
```
fritzbrand, line 576: I don't know where this change came from, I don't recall making it...

home/models.py
```diff
@@ -225,8 +225,14 @@ def clean(self, value):
         result = super().clean(value)
         num_vars_in_msg = len(re.findall(r"{{\d+}}", result["message"]))
         errors = {}
+        example_values = result["example_values"]
+        for ev in example_values:
+            if ev.find(",") != -1:
```
jerith, line 230: Is `ev` a string? Could we perhaps use `if "," in ev:` instead?

home/tests/test_content_import_export.py
jerith: This is a new test that I added. My guess is that you had unsaved changes when you merged in `main` and your editor overwrote the new stuff with the old. Maybe copy this whole file over from `main`?

Either way, this needs to be fixed before merging.

jerith (changes requested):

jerith (approved):

PR #207, opened 2023-10-25
Remove cache from list endpoints.

rudigiesler:
## Purpose
This cache was causing issues with the list endpoints being out of date, since it was set with a very long TTL.

## Approach
There's no way to invalidate this cache on changes, since the key is derived from query string parameters, and Accept and Cookie headers in the request, things that we don't have access to when a page is being changed on the admin.

If it's needed in the future, we will add it back, either with a very short timeout, or a more direct cache that will cache only the things we need, with cache keys that we can delete on changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,9 @@ The current LTS release is: 1.0
 
 ### Redis
 
+For development environments, Redis is optional, it will use an in-memory cache by
+default if the environment variable isn't specified.
+
 For Linux and WSL2 (Windows Subsystem for Linux)
  Creating a docker container with the ports exposed, called `cr_redis`
 `docker run -d -p 6379:6379 --name cr_redis redis`
@@ -60,7 +63,7 @@ For Linux and WSL2 (Windows Subsystem for Linux)
 To then run the docker container,
 `docker run cr_redis`
 
-This can work for mac and (possibly Windows) by setting the environment variable `DATABASE_URL=postgres://postgres@0.0.0.0/contentrepo`
+This can work for mac and (possibly Windows) by setting the environment variable `CACHE=redis://0.0.0.0:6379/0`
 
 ### Postgres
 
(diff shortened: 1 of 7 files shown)


home/tests/test_api.py
```diff
@@ -242,7 +242,7 @@ def test_pagination(self):
         self.assertEqual(content["body"]["text"]["value"]["message"], "WA Message 11")
 
     def test_number_of_queries(self):
-        with self.assertNumQueries(14):
+        with self.assertNumQueries(8):
```
rudigiesler, line 245: This reduced due to in-memory cache being enabled for tests. Wagtail does some caching internally, these additional cached queries aren't anything that we do in our views.

README.md
```diff
@@ -53,14 +53,17 @@ The current LTS release is: 1.0
...
+
 For Linux and WSL2 (Windows Subsystem for Linux)
  Creating a docker container with the ports exposed, called `cr_redis`
 `docker run -d -p 6379:6379 --name cr_redis redis`
 
 To then run the docker container,
 `docker run cr_redis`
 
-This can work for mac and (possibly Windows) by setting the environment variable `DATABASE_URL=postgres://postgres@0.0.0.0/contentrepo`
+This can work for mac and (possibly Windows) by setting the environment variable `CACHE=redis://0.0.0.0:6379/0`
```
jerith, line 66: Should this be `CACHE_URL` instead of `CACHE`?

contentrepo/settings/dev.py
```diff
@@ -16,10 +16,6 @@
...
-# NOTE: We don't want the cache getting in the way during tests, but this also
-# means we're not testing the cache behaviour.
-CACHES = {
-    "default": {
-        "BACKEND": "django.core.cache.backends.dummy.DummyCache",
-    }
-}
+# We default to using local memory for dev and tests, so that Redis isn't a dependancy
+# We test against Redis in the CI
+CACHES = {"default": env.cache("CACHE_URL", default="locmemcache://")}
```
jerith, line 21: Do we really want in-memory caching in tests?

rudigiesler: I thought it might be useful, as that's how we originally picked up that the cache wasn't being invalidated on changes, by a test failing. Now that those endpoints are no longer cached, that issue is fixed, so I thought we could turn it back on again. If we do end up adding caching back again, then we'll want it on, so that we can test that our caching is working correctly.

Happy either way though. I've got the CI testing against redis, to better simulate the production environment, but if we want it off for tests we must just remove redis there as well.

jerith: Ah, right. The internal caching (which is mostly ORM-level, I guess?) is safe -- it's just the HTTP response caching that bit us. Happy with this.

jerith (approved):

PR #208, opened 2023-10-25
Refactor/rewrite content exporter

jerith:
## Purpose
Similar to the import code, the old export code was complicated and fragile. This is a rewrite to make it easier to understand and maintain.

## Approach
Keeping all the existing tests, rewrite the exporter such that all the tests pass. This means that CSV output is equivalent to the old exporter and both CSV and XLSX output passes round-trip export->import tests.

### DIFF
--- a/home/export_content_pages.py
+++ b/home/export_content_pages.py
@@ -75,7 +75,7 @@ class ExportRow:
     def to_tuple(self) -> tuple[str | int, ...]:
         return astuple(self)
 
-    def new_message_row(self, **fields: dict[str, str]) -> "ExportRow":
+    def new_message_row(self) -> "ExportRow":
         return ExportRow(message=self.message + 1, page_id=self.page_id, slug=self.slug)
 
     def new_variation_row(self, variation: VariationBlock) -> "ExportRow":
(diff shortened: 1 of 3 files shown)


fritzbrand (dismissed):

home/export_content_pages.py
```diff
@@ -0,0 +1,372 @@
...
+    def headings(cls) -> list[str]:
+        return [f.name for f in fields(cls)]
+
+    def to_dict(self) -> dict[str, str | int]:
+        return asdict(self)
+
+    def to_tuple(self) -> tuple[str | int, ...]:
+        return astuple(self)
+
+    def new_message_row(self, **fields: dict[str, str]) -> "ExportRow":
```
rudigiesler, line 78: Why do we take in a `fields` parameter here, but never use it?

rudigiesler (approved):

home/export_content_pages.py
jerith: Leftover from a previous iteration of the code. I'll remove it, thanks.

rudigiesler (approved):

PR #211, opened 2023-11-01
Test speedup

rudigiesler:
## Purpose
The tests take quite a long time to run, there seem to be some thing we can do to speed it up, which should speed up iterative development time.

## Approach
We started with a baseline. On my machine the tests took 5:41 to run.

First step was to switch to an in-memory sqlite database for unit tests (CI will still run on postgresql). This brought test run time down to 4:52, or a 14.4% improvement.

Then I switched the API tests from using the importer to populate the database to using the PageBuilder, which is not only easier to understand the fixtures when reading the tests, is also quicker. That brought the time down to 3:30, or a 24% improvement.

I then enabled parallel test execution using `pytest-xdist`. This brought the time down to 1:42, or a 31.7% improvement.

For the tests, I changed the password hasher to MD5, this brought the time down to 1:29, or a 3.8% improvement (possibly larger relative to the others, since this was done after implementing the parallel test execution)

Then I looked at the startup time. Running `pytest --collect-only` took 16.1s. Moving the import of `sentence_transformers` into the if statement means it doesn't get imported for our tests anymore, which reduce startup time to 7.8s, or 51.6% improvement.

The next largest thing to tackle is the migrations, these take quite a long time. `pytest` has a `--no-migrations` flag, which doesn't work immediately because we create data in our migrations (namely the `HomePage`). I've tried adding this as an autouse fixture for all tests, but then get errors relating to the `django_db` mark missing from tests. I'm not sure if this is related to some tests still using the django test classes, instead of pytest things, or some other issue. Either way this seems like a bigger issue, so I won't be addressing it in this PR, but it's a good candidate for future test speedups.

## Learning
- [pytest-profiling](https://pypi.org/project/pytest-profiling/) can give you profiling output for your test runs, which can help you narrow down why tests are slow.
- The `--durations=10` option for `pytest` will give you the 10 tests that take the longest. Right now the top 4 items there are test setup (the name of the test listed there is irrelevant), one for each process running tests in parallel.
- You can use `python -m cProfile -o profile -m pytest --collect-only` to profile the startup time, and then use `flameprof profile > output.svg` to get a flamegraph of the test startup time.


fritzbrand (approved):

jerith:
I prefer to avoid `pytest-xdist` unless there's a very specific need for it, because it has implications for any fixtures scoped above function-level. Now that we're using in-memory sqlite that should be okay, but it's probably worth checking if it's actually giving us a meaningful benefit after the other changes. (And if the speedup is small, maybe we should leave it out for now.)

rudigiesler:
@jerith Happy to remove it (or remove it from default and just run it locally, keeping CI non-parallel since I most of the CI time is spent installing large dependancies anyway), but it gave me quite a substantial speedup. Went from 3:30 (that's with the in-memory sqlite) to 1:42, so quite a substantial speedup, and that's on my machine where I only have 2 cores + 2 threads. It does mean that the test setup in test_api is now done once per core instead of once, but even with that we still have a significant speedup.

PR #214, opened 2023-11-09
Test improvements (part 1)

jerith:
## Purpose
We don't have great test coverage of all the features, and improving our tests will likely uncover implicit assumptions that will inform the design work we need to do.

## Approach
Go through test modules ones at a time, figuring out what the tests do and looking for ways to improve them. Since there's a lot to go through, splitting it into separate PRs seems reasonable.

### DIFF
--- a/home/tests/test_api.py
+++ b/home/tests/test_api.py
@@ -26,47 +26,52 @@ def uclient(client, django_user_model):
     return client
 
 
-@pytest.fixture()
-def pagination_test_data():
-    home_page = HomePage.objects.first()
-    main_menu = PageBuilder.build_cpi(home_page, "main-menu", "Main Menu")
-    content_page1 = PageBuilder.build_cp(
-        parent=main_menu,
-        slug="main-menu-first-time-user",
-        title="main menu first time user",
-        bodies=[
-            WABody("main menu first time user", [WABlk("*Welcome to HealthAlert* 🌍")]),
-            MBody("main menu first time user", [MBlk("*Welcome to HealthAlert* 🌍")]),
-        ],
-        tags=["menu"],
-        quick_replies=["Self-help", "Settings", "Health Info"],
-        triggers=["Main menu"],
-    )
-    PageBuilder.build_cp(
-        parent=content_page1,
-        slug="health-info",
-        title="health info",
-        bodies=[
-            WABody("health info", [WABlk("*Health information* 🏥")]),
-            MBody("health info", [MBlk("*Health information* 🏥")]),
-        ],
-        tags=["health_info"],
-    )
-    PageBuilder.build_cp(
-        parent=content_page1,
-        slug="self-help",
-        title="self-help",
-        bodies=[
-            WABody("self-help", [WABlk("*Self-help programs* 🌬️")]),
-            MBody("self-help", [MBlk("*Self-help programs* 🌬️")]),
-        ],
-        tags=["self_help"],
-    )
-
-
-@pytest.mark.usefixtures("pagination_test_data")
 @pytest.mark.django_db
 class TestContentPageAPI:
+    @pytest.fixture(autouse=True)
+    def create_test_data(self):
+        """
+        Create the content that all the tests in this class will use.
+        """
+        home_page = HomePage.objects.first()
+        main_menu = PageBuilder.build_cpi(home_page, "main-menu", "Main Menu")
+        content_page1 = PageBuilder.build_cp(
+            parent=main_menu,
+            slug="main-menu-first-time-user",
+            title="main menu first time user",
+            bodies=[
+                WABody(
+                    "main menu first time user", [WABlk("*Welcome to HealthAlert* 🌍")]
+                ),
+                MBody(
+                    "main menu first time user", [MBlk("*Welcome to HealthAlert* 🌍")]
+                ),
+            ],
+            tags=["menu"],
+            quick_replies=["Self-help", "Settings", "Health Info"],
+            triggers=["Main menu"],
+        )
+        PageBuilder.build_cp(
+            parent=content_page1,
+            slug="health-info",
+            title="health info",
+            bodies=[
+                WABody("health info", [WABlk("*Health information* 🏥")]),
+                MBody("health info", [MBlk("*Health information* 🏥")]),
+            ],
+            tags=["health_info"],
+        )
+        PageBuilder.build_cp(
+            parent=content_page1,
+            slug="self-help",
+            title="self-help",
+            bodies=[
+                WABody("self-help", [WABlk("*Self-help programs* 🌬️")]),
+                MBody("self-help", [MBlk("*Self-help programs* 🌬️")]),
+            ],
+            tags=["self_help"],
+        )
+
     def test_login_required(self, client):
         """
         Users that aren't logged in shouldn't be allowed to access the API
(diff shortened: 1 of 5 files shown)


home/tests/test_api.py
```diff
@@ -2,234 +2,244 @@
...
-    def test_login_required(self):
+@pytest.fixture()
+def uclient(client, django_user_model):
+    creds = {"username": "test", "password": "test"}
+    django_user_model.objects.create_user(**creds)
+    client.login(**creds)
+    return client
+
+
+@pytest.fixture()
```
rudigiesler, line 29: Do we need to set the scope of this fixture, to avoid it being recreated every test, or is the default scope fine?

home/tests/test_api.py
```diff
@@ -2,234 +2,244 @@
...
-        self.assertEqual(content["body"]["text"]["message"].replace("\r", ""), message)
+        assert not page2.live
+        assert content["body"]["text"]["message"].replace("\r", "") == message
+
+    def test_pagination(self, uclient):
+        """
+        FIXME:
+         * It's unclear what this is actually testing.
+         * Should it be multiple tests instead of just one?
+        """
```
rudigiesler, line 201: This looks to be testing getting multiple whatsapp messages for a single ContentPage, the `?message=...` query parameter.

I think it could definitely be split out into something like: whatsapp body when whatsapp disabled, first message when no `message` parameter specified, error if message index is greater than number of messages, error if message index is not an integer, and specific message if message query param is supplied.

home/tests/test_api.py
```diff
@@ -244,118 +254,142 @@ def test_pagination(self):
...
 
-    def test_number_of_queries(self):
-        with self.assertNumQueries(8):
-            self.client.get("/api/v2/pages/")
+    def test_number_of_queries(self, uclient, django_assert_num_queries):
+        """
+        FIXME:
+         * Is the number of queries significant? What does it depend on?
+         * Should this be part of some other test?
+        """
```
rudigiesler, line 274: I think the point of this is to ensure that we're not doing n+1 queries when listing the pages.

Usually for this kind of test, I run it to assert on an incorrect number of queries, which gives me a list of queries, that I can investigate. Then I leave a comment with a description of each of the queries, so that by looking at the test, someone can easily tell what each of the queries are.

home/tests/test_api.py
```diff
@@ -244,118 +254,142 @@ def test_pagination(self):
...
+            uclient.get("/api/v2/pages/")
 
-    def test_detail_view(self):
+    def test_detail_view(self, uclient):
+        """
+        FIXME:
+         * It's unclear what this is actually testing.
+         * Should it be multiple tests instead of just one?
+         * This should probably be in a class without the content setup.
+        """
```
rudigiesler, line 284: I think this is testing the format of the `/pages/{page_id}/` endpoint. I think this does need to be multiple tests, since there's a bit of `?whatsapp=true` in there, which is a bit odd, that bit can be a separate test, I think here we just want to test the bits of the response that are the same across all channel types. It also tests that page views are created on a `GET` to the page.

We could have it as a separate class, or use one of the ContentPages we've already created in the setup, instead of deleting and creating a new one.

home/tests/test_api.py
```diff
@@ -244,118 +254,142 @@ def test_pagination(self):
...
-        self.assertEqual(view.message, 1)
+        assert view.message == 1
 
-    def test_detail_view_with_variations(self):
+    def test_detail_view_with_variations(self, uclient):
+        """
+        FIXME:
+         * It's unclear what this is actually testing.
+         * This should probably be in a class without the content setup.
+        """
```
rudigiesler, line 328: This is intended to test the `variation_messages` key in the API result.

home/tests/test_api.py
```diff
@@ -244,118 +254,142 @@ def test_pagination(self):
...
+        assert content["body"]["is_whatsapp_template"]
+        assert content["body"]["whatsapp_template_name"] == "test_template"
+        assert content["body"]["text"]["value"]["message"] == "Test WhatsApp Message 1"
 
-    def test_detail_view_no_content_page(self):
+    def test_detail_view_no_content_page(self, uclient):
+        """
+        FIXME:
+         * It's unclear what the intended result is.
+         * Should the response code be a success?
```
rudigiesler, line 372: I'm unsure whether it makes sense to be a success or not. I've seen some good arguments not to return a 404 in REST APIs if an object with that key doesn't exist, because that doesn't tell you if you've got the URL path wrong, or if it's an object that doesn't exist.

Either way, we should probably assert on the response code to make it clear what we expect.

home/tests/test_api.py
```diff
@@ -244,118 +254,142 @@ def test_pagination(self):
...
-    def setUp(self):
-        self.user_credentials = {"username": "test", "password": "test"}
-        self.user = get_user_model().objects.create_user(**self.user_credentials)
-        self.client.login(**self.user_credentials)
+@pytest.mark.django_db
+class TestWhatsAppMessages:
+    """
+    FIXME:
+     * Should some of the WhatsApp tests from TestPagination live here instead?
+    """
```
rudigiesler, line 387: Yes, I think it's probably a good idea to move tests like `test_whatsapp_body` in here

home/tests/test_api.py
```diff
@@ -376,15 +410,17 @@ def test_whatsapp_detail_view_with_button(self):
...
-        self.assertEqual(
-            button, {"type": "next_message", "value": {"title": "Tell me more"}}
-        )
+        assert button == {"type": "next_message", "value": {"title": "Tell me more"}}
 
-    def test_whatsapp_template(self):
+    def test_whatsapp_template(self, uclient):
+        """
+        FIXME:
+         * Is this actually a template message?
```
rudigiesler, line 422: Yeah I think there's a boolean field here that also needs to be set, `is_whatsapp_template`, as well as `whatsapp_template_name`

home/tests/test_api.py
jerith: Each test function runs in its own db transaction (as with Django's `TestCase`), so we need to populate the db for every test. (We were doing that with the `setUp()` methods and such anyway, so there's no real change here.)

home/tests/test_api.py
jerith: 🤔

home/tests/test_api.py
jerith: 🤔

home/tests/test_api.py
jerith: Since variations are (currently) WhatsApp-only, should I move this to `TestWhatsAppMessages`?

home/tests/test_api.py
jerith: Proper REST APIs should be giving you resource URLs instead of making you construct them yourself, so getting the structure wrong shouldn't be a concern there. 🤷 I think it's safe to assume that the client is either going to be requesting correct URLs (that may point to things that don't exist) in which case 404 for "the thing you asked for isn't here" makes sense, or it's so broken that differentiating between "not found" and "nonsensical request" won't actually help at all.

I'll add an assert on the response code and maybe a comment if it's not 404.

home/tests/test_api.py
jerith: 🤔

home/tests/test_api.py
```diff
@@ -2,234 +2,244 @@
...
+    )
+
+
+@pytest.mark.usefixtures("pagination_test_data")
+@pytest.mark.django_db
+class TestPagination:
+    """
+    FIXME:
+     * Most of these tests don't seem to have anything to do with pagination.
+    """
```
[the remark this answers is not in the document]
jerith: I'll rename `TestOrderedContentSet` to `TestOrderedContentSetAPI` as well, for consistency.

home/tests/test_api.py
jerith: Ugh. This does two extra queries when it's the first test that runs.

jerith: I have no idea what all these queries are doing, so I haven't tried to document them. 🤷

home/tests/test_api.py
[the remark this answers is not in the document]
jerith: Done.

home/tests/test_api.py
jerith: Done.

home/tests/test_api.py
jerith: Looks like we explicitly catch `ContentPage.DoesNotExist` (presumably 404) and raise `ValidationError` (400) instead. This definitely needs a rethink. 😕

home/tests/test_api.py
jerith: Done.

home/tests/test_api.py
rudigiesler: The `setUpTestData` method did it once per class, and then ran all the tests against that setup state, I assume with transaction checkpoints or something.

home/tests/test_api.py
rudigiesler: Ah, I imagine that's internal wagtail things. This used to take a lot more queries before we enabled cache for testing, so if you want fix the "different number of queries if run first" issue you could disable cache for just this test and update the number of queries.

home/tests/test_api.py
rudigiesler: Hmm, looks to be that we don't want a 500 error on missing IDs, which is correct, and would happen if we try to save a page view for a ContentPage that doesn't exist. But then we avoid that by raising a `ValidationError`, which is a 400 error, which doesn't seem correct to me..

home/tests/test_api.py
jerith: Ah, I forgot about that. It doesn't seem to make the tests any slower, so I don't think it's a problem.

home/tests/test_api.py
jerith: Making the request twice and only counting the second sorts out the variance.  We probably want to measure this with caching enabled anyway, so I'll leave it as-is.

home/tests/test_api.py
jerith: We can raise `rest_framework.exceptions.NotFound` instead of `rest_framework.exceptions.ValidationError`, but that's an API change. Something for the design doc or a separate issue, I guess. 🤷

jerith:
Based on coverage numbers, we don't have any tests for triggers or whatever the `s` query param is (looks like something to do with word embeddings?) -- we probably want to add those a separate PR. I have no idea what those features even are, so it's probably best if someone else drives that. (I'm happy to help, though.)

rudigiesler:
I would assume that triggers has to do with filtering results based on whether the contentpage has a specific trigger in it's `trigger` field. I'm not sure how the word embeddings work though...

home/tests/test_api.py
jerith: I've split this into four separate tests with more thorough assertions that use the existing page data instead of replacing it with new data.

rudigiesler (approved):

home/tests/test_api.py
jerith: I'm going to leave this as-is for now.

home/tests/test_api.py
jerith: I'm going to leave this as-is for now.

PR #215, opened 2023-11-16
WhatsApp template language support

jerith:
## Purpose
Hardcoding `en_US` for all WhatsApp templates isn't ideal for non-English content. We should use the relevant locale instead.

## Approach
Look up the closest WhatsApp supported language code for the page locale and use that instead of a hardcoded `en_US`.


Buhle79:
Hi @jerith I've resolved the conflicts and merged with main.

home/tests/test_whatsapp.py
```diff
@@ -166,10 +187,47 @@ def test_create_whatsapp_template_with_image(self, tmp_path, settings):
...
+    def test_create_whatsapp_template_with_language(self) -> None:
+        """
+        When we create a WhatsApp template, the language is converted to the
+        appropriate WhatsApp language code.
+        """
+        url = "http://whatsapp/graph/v14.0/27121231234/message_templates"
+        responses.add(responses.POST, url, json={})
+
+        # The default English locale gives us a language of "en_US".
+        # FIXME: Should this be "en" instead?
```
rudigiesler, line 203: It looks like we can define both in our settings file: https://docs.wagtail.org/en/stable/reference/settings.html#wagtail-content-languages

Maybe we should ensure that that list matches what we can do with WhatsApp templates? We probably don't want to remove any languages from the list, but we can do something like have all 3 of `en`, `en-us`, `en-gb`, and then for backwards compatibility add a migration that changes any `en` locales to `en-us`

home/whatsapp.py
```diff
@@ -1,15 +1,121 @@
...
+    category: str,
+    locale: Locale | None = None,
+    quick_replies: Iterable[str] = (),
+    image_id: int | None = None,
+    example_values: Iterable[str] | None = None,
+) -> None:
+    """
+    Create a WhatsApp template through the WhatsApp Business API.
+
+    FIXME: Do we want locale to be optional?
```
rudigiesler, line 114: The language field is required for the template API, so I would say that we want locale to be required here too.

rudigiesler (commented):
Trying to answer all the TODOs and fixme's here. I think I've come up with a way of supporting all 3 of `en`, `en-us`, `en-gb`, while keeping backwards compat, let me know what you think.

home/tests/test_whatsapp.py
jerith: This (somewhat poorly worded, sorry) FIXME was about switching the default WhatsApp language for wagtail's `en` from the current `en_US` to `en`. I think we've already decided to keep that `en_US` for backwards compatibility with existing templates.

home/tests/test_models.py
```diff
@@ -79,11 +81,12 @@ def test_template_create_on_save_deactivated(self, mock_create_whatsapp_template
     @mock.patch("home.models.create_whatsapp_template")
     def test_template_create_on_save(self, mock_create_whatsapp_template):
         page = create_page(is_whatsapp_template=True)
+        locale = Locale.objects.get(language_code="en")
```
jerith, line 84: Can we call this `en` rather than `locale`, please?

home/tests/test_models.py
```diff
@@ -358,6 +368,25 @@ def test_for_missing_migrations(self):
             "There are missing migrations:\n %s" % output.getvalue(),
         )
 
+    @override_settings(WHATSAPP_CREATE_TEMPLATES=True)
+    @mock.patch("home.models.create_whatsapp_template")
+    @pytest.mark.xfail(
+        reason="This fails because we can't get locale to create the page, "
+        "these tests will be changed once whatsapp templates are separated."
+    )
```
jerith, line 376: Note for other reviewers: We're replacing these tests with better ones (that don't use mocks, to start with) in #277. It's not worth duplicating that work here.

fritzbrand (approved):
Looks good to me

PR #216, opened 2023-12-04
Fix importing of multi-language content (again)

jerith:
## Purpose
Now that we've decided that slugs are unique per-locale rather than globally unique, the importer needs to identify pages by (slug, locale) rather than just slug.


home/import_content_pages.py
```diff
@@ -57,11 +47,25 @@ def __init__(
...
         self.locale = locale
-        self.shadow_pages: dict[str, ShadowContentPage] = {}
+        self.locale_map: dict[str, Locale] = {}
+        self.shadow_pages: dict[PageId, ShadowContentPage] = {}
         self.go_to_page_buttons: dict[
-            str, dict[int, list[dict[str, Any]]]
+            PageId, dict[int, list[dict[str, Any]]]
         ] = defaultdict(lambda: defaultdict(list))
 
+    def locale_from_display_name(self, langname: str) -> str:
```
rudigiesler, line 56: Isn't this returning a `Locale`? And is there a reason for managing our own map vs using the `cache` decorator?

rudigiesler (commented):

rudigiesler (approved):

PR #222, opened 2023-12-20
Import error propagation

rudigiesler:
## Purpose
If there are any issues during import, currently the admin user gets a generic error, we roll back, and log the error message.

Someone needs to then go and investigate the error message, figure out what went wrong with the import, and fix it.

Often there are issues with the import file that are not an issue with the import code. In those cases, we want to rather show an error message to the admin user with enough information for them to fix the import file, rather than logging an error.

## Approach
We create a new exception, ImportException, and if we have an exception of this error type, we rather put the error information in the error message sent back to the admin user, instead of using the generic error and logging the exception. This will hopefully give the admin user enough information to fix the issue in the import file, and try again.

image

### DIFF
--- a/home/import_content_pages.py
+++ b/home/import_content_pages.py
@@ -297,7 +297,6 @@ class ContentImporter:
             page.enable_whatsapp = True
             buttons = []
             for button in row.buttons:
-                print(button)
                 if button["type"] == "next_message":
                     buttons.append(
                         {
@@ -431,7 +430,16 @@ class ShadowContentPage:
         page = ContentPage.objects.get(slug=self.slug, locale=self.locale)
         related_pages = []
         for related_page_slug in self.related_pages:
-            related_page = Page.objects.get(slug=related_page_slug, locale=self.locale)
+            try:
+                related_page = Page.objects.get(
+                    slug=related_page_slug, locale=self.locale
+                )
+            except Page.DoesNotExist:
+                raise ImportException(
+                    f"Cannot find related page with slug {related_page_slug} and "
+                    f"locale {self.locale}",
+                    self.row_num,
+                )
             related_pages.append(("related_page", related_page))
         page.related_pages = related_pages
         page.save_revision().publish()
--- a/home/import_content_pages.py
+++ b/home/import_content_pages.py
@@ -97,7 +97,7 @@ class ContentImporter:
         # Non-page rows don't have a locale, so we need to remember the last
         # row that does have a locale.
         prev_locale: Locale | None = None
-        for i, row in enumerate(rows):
+        for i, row in enumerate(rows, start=2):
             try:
                 if row.is_page_index:
                     if self.locale and row.locale != self.locale.get_display_name():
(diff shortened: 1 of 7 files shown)


home/import_content_pages.py
```diff
@@ -84,20 +93,24 @@ def process_rows(self, rows: list["ContentRow"]) -> None:
...
+                    prev_locale = self.locale_from_display_name(row.locale)
+                elif row.is_content_page:
+                    self.create_shadow_content_page_from_row(row)
+                    prev_locale = self.locale_from_display_name(row.locale)
+                elif row.is_variation_message:
+                    self.add_variation_to_shadow_content_page_from_row(row, prev_locale)
+                else:
+                    self.add_message_to_shadow_content_page_from_row(row, prev_locale)
+            except ImportException as e:
+                e.row_num = i + 2  # Rows numbers start at one and have a header row
```
jerith, line 112: You could use `enumerate(rows, 2)` above instead of adding 2 where you use the index.

I think the add is fine in this case, because we're only using the index once, `enumerate()`'s second arg can be useful when the index is referenced in a bunch of different places.

jerith (commented):
The CI failure is from the linter. Make sure you're using the newest black/ruff/mypy versions locally.

home/tests/test_content_import_export.py
```diff
@@ -13,12 +13,14 @@
 from django.core import serializers  # type: ignore
 from django.core.exceptions import ValidationError  # type: ignore
 from django.core.files.images import ImageFile  # type: ignore
+from django.test import override_settings  # type: ignore
```
jerith, line 16: pytest-django has a `settings` fixture that we're already using in our own `tmp_media_path` fixture.

jerith (commented):
Do we want a test for the got-to-page button with a missing page?

Also, I think we (silently) ignore related pages we can't find. Should that be an `ImportError` as well?

home/import_content_pages.py
```diff
@@ -295,6 +297,7 @@ def add_message_to_shadow_content_page_from_row(
             page.enable_whatsapp = True
             buttons = []
             for button in row.buttons:
+                print(button)
```
jerith, line 300: Stray print?

home/tests/test_content_import_export.py
```diff
@@ -858,6 +858,21 @@ def test_multiple_parents(self, newcsv_impexp: ImportExport) -> None:
...
+        Go to page buttons in the import file link to other pages using the slug. But
+        if no page with that slug exists, then we should give the user an error message
+        that tells them where and how to fix it.
+        """
+        with pytest.raises(ImportException) as e:
+            newcsv_impexp.import_file("missing-gotopage.csv")
+        assert e.value.row_num == 2
+        assert (
+            e.value.message
+            == "No pages found with slug missing and locale English for go_to_page "
```
jerith, line 872: "No pages found with slug missing" could mean that the slug is missing, rather than having the value `missing`. Do we maybe want to quote it in the error message? Alternatively, using a different value (perhaps `missing-page`) in the test would make it clearer here.

jerith (commented):
Perhaps for a later PR: We're accumulating quite a collections of test CSVs. Should we move them all into a subdir to reduce clutter in the test dir?

jerith (approved):
🎉

PR #223, opened 2024-01-03
Mnch 985 contentrepo whatsapp template

Buhle79:
## Purpose
Template was store with uppercase and throw an error when they try to retrieve it as template names are stored with lower case.

## Approach
Convert whatsapp template name to lower case before we store it.

## Learning
_Describe the research stage_

Create a whatsapp template locally and modify unittests

None

#### Blog Posts

### DIFF
--- a/home/models.py
+++ b/home/models.py
@@ -605,9 +605,7 @@ class ContentPage(UniqueSlugMixin, Page, ContentImportMixin):
 
     @property
     def whatsapp_template_prefix(self):
-        if self.whatsapp_title:
-            self.whatsapp_title = self.whatsapp_title.lower()
-        return self.whatsapp_title.replace(" ", "_")
+        return self.whatsapp_title.lower().replace(" ", "_")
 
     @property
     def whatsapp_template_body(self):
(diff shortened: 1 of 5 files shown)


Buhle79:
Hi @jerith for review please

home/models.py
```diff
@@ -605,6 +605,8 @@ def latest_revision_rating(self):
 
     @property
     def whatsapp_template_prefix(self):
+        if self.whatsapp_title:
+            self.whatsapp_title = self.whatsapp_title.lower()
```
jerith, line 609: Do we want to set `self.whatsapp_title` to lowercase here, or just return a lowercase version of it?

Also, on the line below we call `.replace()` on it unconditionally. That means we don't need to check if it's non-empty before lowercasing it.

rudigiesler (approved):
Looks good!

I think we might want a data migration, to lowercase all the existing template names that are in the database, see https://github.com/praekeltfoundation/contentrepo/blob/main/home/migrations/0030_contentpage_whatsapp_template_name.py for an example, but happy for that to happen in a separate PR.

PR #224, opened 2024-01-08
Mnch 985 contentrepo whatsapp template

Buhle79:
## Purpose
Fix historic data for template name to lowercase.

## Approach
Add migration script

## Learning
_Describe the research stage_

#### Blog Posts


Buhle79:
Hi @rudigiesler for review please.

home/migrations/0041_contentpage_whatsapp_template_lower_case_name.py
```diff
@@ -0,0 +1,46 @@
...
+
+def update_template_names(ContentPage, ContentType, Revision):
+    """
+     Update historic data template names to lower case
+    """
+    for page in ContentPage.objects.filter(is_whatsapp_template=True):
+        if not page.whatsapp_title:
+            continue
+        template_prefix = page.whatsapp_title.lower().replace(" ", "_")
+        page.whatsapp_template_name = f"{template_prefix}_{page.latest_revision.id}"
```
rudigiesler, line 14: I don't think we need to rename the templates, in fact that might lead to issues since they're already submitted to whatsapp. I think we just need to do what the `submit_whatsapp_template` function did, which was lowercase them, ie. `page.whatsapp_template_name = page.whatsapp_template_name.lower()`. (and the same for revisions).

I think then we'll also need to change the `if not page.whatsapp_title` check, since we now need to check if there's a `page.whatsapp_template_name`

home/migrations/0041_contentpage_whatsapp_template_lower_case_name.py
```diff
@@ -0,0 +1,46 @@
...
+        ("home", "0040_alter_contentpage_whatsapp_template_category"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="contentpage",
+            name="whatsapp_template_name",
+            field=models.CharField(max_length=512, blank=True, default=""),
+        ),
+    ]
```
rudigiesler, line 46: I think this `operations` here isn't correct. I don't think we're altering the `whatsapp_template_name` field at all, just the contents of the field, so we rather want a `RunPython` migration here, that will call the `run_migration` function. At the moment I don't think the migration is getting called.

rudigiesler (commented):
Just a comment on what we're changing here. I think we just need to lowercase the existing template names in the database, not regenerate them. Otherwise looks good! I like the approach

home/migrations/0041_contentpage_whatsapp_template_lower_case_name.py
Buhle79: Can we please chat about this after SU.

PR #225, opened 2024-01-11
confidance threshold to word embedding

TatendaMugadza:
- added confidence threshold for word embedding results
- Improved word embedding preprocessing
- Updated read me for embedding generation

## Purpose
The aaq search was returning results even in the cases when the user has entered meaningless invalid text
https://praekelt.leankit.com/card/31512186572175

## Approach
Added a confidence threshold so that we can filter out any returned results that have a similarity score of less than 25%

## Learning
From analysing how different search inputs compare to the results, in terms of similarity score, 25% similarity gave a balance between giving back relevant answers and filtering out completely unrelated responses

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,7 @@ Please make sure to update tests as appropriate.
 
 ## Content Embeddings
 if you want to generate content content embeddings, please make sure you define LOAD_TRANSFORMER_MODEL in your environment and set it to True
+NB: this causes the model to be loaded into memory and can significantly increase resource requirements
 
 ## License
 [MIT](https://choosealicense.com/licenses/mit/)
--- a/README.md
+++ b/README.md
@@ -128,5 +128,8 @@ Pull requests are welcome. For major changes, please open an issue first to disc
 
 Please make sure to update tests as appropriate.
 
+## Content Embeddings
+if you want to generate content content embeddings, please make sure you define LOAD_TRANSFORMER_MODEL in your environment and set it to True
+
 ## License
 [MIT](https://choosealicense.com/licenses/mit/)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -128,5 +128,8 @@ Pull requests are welcome. For major changes, please open an issue first to disc
 
 Please make sure to update tests as appropriate.
 
+## Content Embeddings
+if you want to generate content content embeddings, please make sure you define LOAD_TRANSFORMER_MODEL in your environment and set it to True
```
KaitCrawford, line 132: I think adding something like
"Note: this causes the model to be loaded into memory and can significantly increase resource requirements"

home/tests/test_api.py
```diff
@@ -110,6 +110,32 @@ def test_tag_filtering(self, uclient):
...
+        content = json.loads(response.content)
+        assert content["count"] == 0
+        # it should return 0 pages for correct search term if no platform is provided
+        response = uclient.get("/api/v2/pages/?s=help")
+        content = json.loads(response.content)
+        print("content is ",content)
+        assert content["count"] == 0
+        # it should not return search term matching pages if they are unpublished
+        page1.unpublish()
+        response = uclient.get("/api/v2/pages/?s=#(&whatsapp=true")
```
KaitCrawford, line 135: This is the same search term that previously didn't return any results. You probably want to use `s=help` here for the test to be valid

KaitCrawford (approved):
@TatendaMugadza two small comments but it looks good 👍

README.md
TatendaMugadza: This has been updated

home/tests/test_api.py
rudigiesler: Just to add to this, I don't think we should be using the `#` character in the URL for our tests for this. It's the fragment identifier character, and actually needs to be escaped if you want to use it in URL parameters, otherwise it could result in everything after the `#` being interpreted as a fragment identifier and not as part of the query string: https://www.rfc-editor.org/rfc/rfc3986.html#section-3.5

PR #228, opened 2024-01-16
valid-variable-check

Mudiwa66:
## Purpose
When users create templates that need to be submitted to Turn, they need to input valid variables only. We need to add a check for valid variables, according to the guidelines from Meta.

## Approach
User's will now only be able to input valid variables when creating and submitting a template to turn. Specific errors will also be presented.


home/models.py
```diff
@@ -295,7 +295,29 @@ def clean(self, value):
                 f"{self.MEDIA_CAPTION_MAX_LENGTH} characters, your message is "
                 f"{len(result['message'])} characters long"
             )
+
+        # find variables
+        vars_in_msg = re.findall(r"{{(.*?)}}", result["message"])
+        non_digit_variables = [var for var in vars_in_msg if not var.isdigit()]
```
rudigiesler, line 301: I think `isdigit` will match on things we don't want, like superscripts and Kharosthi numbers: https://docs.python.org/3/library/stdtypes.html#str.isdigit . I don't think WhatsApp templates support those.

Maybe we want `isdecimal` instead? https://docs.python.org/3/library/stdtypes.html#str.isdecimal

home/models.py
```diff
@@ -295,7 +295,29 @@ def clean(self, value):
...
+        if non_digit_variables != []:
```
rudigiesler, line 302: Usually in Python, we use the fact that lists are truthy if they are not empty, to create an if statement that's easier to read, ie. `if non_digit_variables:`

home/models.py
```diff
@@ -295,7 +295,29 @@ def clean(self, value):
...
+            errors["message"] = ValidationError("Please enter numeric variables only.")
```
rudigiesler, line 303: Since we've got a list of the variables that are non-numeric, `non_digit_variables`, maybe we can include that in the error message to make it easier for the user to find out what we're referring to. We can probably do something similar with the other error message as well.

home/models.py
```diff
@@ -295,7 +295,29 @@ def clean(self, value):
...
+        # check variable order
+        digit_variables = [var for var in vars_in_msg if var.isdigit()]
+        if digit_variables != []:
+            if (
+                all(
+                    x < y
+                    for x, y in zip(digit_variables, digit_variables[1:], strict=False)
+                )
+                is not True
+                or digit_variables[0] == "0"
```
rudigiesler, line 313: Instead of having this all in one giant if statement, maybe it's better to separate it out. At the moment it's rather difficult to figure out what this code is doing at a glance.

A cleaner way of doing this might be to create a list of what we expect the values of the variables to be, and then to compare that to the list that we extracted from the message. If they don't equal, then we know we need to raise a validation error. eg:
```
expected_variables = [str(i + 1) for i in range(len(vars_in_msg))]
if vars_in_message != expected_variables:
    errors....
```

home/models.py
```diff
@@ -295,7 +295,29 @@ def clean(self, value):
...
+            ):
+                errors["message"] = ValidationError(
+                    "Variables must be sequential, starting with {{1}}."
+                )
+
         if errors:
+            print(errors)
```
rudigiesler, line 320: We can remove this print before merging, I don't think we need it.

rudigiesler (commented):
Looks good! Thanks :smile: 

The one thing I'm unsure about though, is what happens if the "is_whatsapp_template" checkbox is not checked? Does it still check that the variables are numeric and are in order? 

If this isn't a whatsapp template, then we won't be submitting this to whatsapp, so we don't need to check that the variables conform. For example, we might want a normal message with `Hi {{name}}!`, and then our front-end will replace `{{name}}` with the name of the user. For a template this has to be `Hi {{1}}!`, since that's what WhatsApp enforces, but on our front-end it would be nicer if the variables can tell us what we should fill in them.

I think the difficulty of this is the `is_whatsapp_template` is a field on the page, not on the whatsapp message, so we'll have to do this in the `clean` method of the page, but still make sure that the validation error points to the correct field inside the correct whatsapp message on the page.

home/migrations/0042_variable_validation.py
```diff
@@ -0,0 +1,190 @@
...
+                            help_text="Each message will be sent with the text and media",
+                        ),
+                    )
+                ],
+                blank=True,
+                null=True,
+                use_json_field=True,
+            ),
+        ),
+    ]
```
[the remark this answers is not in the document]
Mudiwa66: You're correct, it's not needed. Should I revert to the previous migration then delete my migration file?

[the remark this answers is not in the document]
rudigiesler: Yes you can. Reverting will help your local database not get in a weird state, and then you can delete the files and commit that change.

home/tests/test_models.py
```diff
@@ -225,16 +250,41 @@ def test_template_not_submitted_with_no_message(
...
-            whatsapp_template_name="WA_Title_1",
-        )
+        with self.assertRaises(ValidationError):
+            PageBuilder.build_cp(
+                parent=main_menu,
+                slug="ha-menu",
+                title="HealthAlert menu",
+                bodies=[],
+                whatsapp_template_name="WA_Title_1",
+            )
```
rudigiesler, line 260: Having a look at this test in more depth, I think we should rather keep it's original functionality. The reason it is now failing is because it doesn't set a title, which we're now checking for. So I think we should rather add a title (change the `bodies=[]` to `bodies=[WABody("WA Title", [])]`), and then remove the `with self.assertRaises(ValidationError):`

That would then ensure that the title validation check passes, and the test should then pass without raising the ValidationError

home/tests/test_models.py
```diff
@@ -225,16 +250,41 @@ def test_template_not_submitted_with_no_message(
...
+                        [
+                            WABlk(
+                                "{{2}}{{1}} {{foo}}",
+                            )
+                        ],
+                    )
+                ],
+                whatsapp_template_name="WA_Title_1",
+            )
+
```
rudigiesler, line 287: I think we also need a test for the WhatsApp title validation, to ensure that if there's no whatsapp title, then the ValidationError is raised to send the error message to the user

home/models.py
```diff
@@ -748,6 +749,61 @@ def save_revision(
...
+                str(j + 1) for j in range(len(actual_digit_variables))
+            ]
+            if actual_digit_variables != expected_variables:
+                errors.setdefault("whatsapp_body", []).append(
+                    StreamBlockValidationError(
+                        {
+                            0: StreamBlockValidationError(
+                                {
+                                    "message": ValidationError(
+                                        f'Variables must be sequential, starting with "{{1}}". Your first variable was "{actual_digit_variables[0]}"'
```
rudigiesler, line 794: Is there a reason why we're only giving the value for the first variable? If I have a message with `{{1}} {{3}} {{2}}`, then the error message I get is `Variables must be sequential, starting with "{1}". Your first variable was "1"`, which isn't very useful. Maybe `Your variables are {actual_digit_variables}`?

rudigiesler (commented):
Looks good! The main thing I think is missing is a test for the new check to ensure that `whatsapp_title` is not empty.

home/models.py
Mudiwa66: Not sure why I put that there, its been removed.

home/tests/test_models.py
Mudiwa66: This new test_template_submitted_with_no_title fails if I don't raise the validation, which is what I should expect. So I added the  `with self.assertRaises(ValidationError):` part so it passes.

rudigiesler (approved):

PR #229, opened 2024-01-16
Add ussd support

fritzbrand:
## Purpose
Add USSD support to ContentRepo


home/models.py
```diff
@@ -300,6 +300,36 @@ def clean(self, value):
...
+        icon = "user"
+        form_classname = "whatsapp-message-block struct-block"
+
+    def clean(self, value):
+        result = super().clean(value)
+        errors = {}
+
+        if errors:
+            raise StructBlockValidationError(errors)
+        return result
```
rudigiesler, line 330: I think we can probably do without this `clean` method? It doesn't seem to do anything from what I can see

home/models.py
```diff
@@ -300,6 +300,36 @@ def clean(self, value):
         return result
 
 
+class SMSBlock(blocks.StructBlock):
+    message = blocks.TextBlock(
+        help_text="each message cannot exceed 160 characters.",
+        validators=(MaxLengthValidator(160),),
+    )
```
rudigiesler, line 307: I think keep it this way for now, but one thing to think of is that we can have multi-part smses, where the message gets split and sent (and charged) as multiple messages, but then the phone reassembles them on the other side and displays it as a single message. Previously I think we've not used this to keep cost low, but just to keep it in mind. I think we can create another block type to handle that though

home/tests/test_content_import_export.py
```diff
@@ -990,8 +1029,11 @@ def test_web_content(self, impexp: ImportExport) -> None:
         )
 
         orig = impexp.get_page_json()
+        print(orig)
         impexp.export_reimport()
         imported = impexp.get_page_json()
+        print("|||")
+        print(imported)
```
rudigiesler, line 1036: I don't think we need these print statements anymore

rudigiesler (commented):
Looks good! The only place I think we would be missing any sort of tests is with the CSV import/export. I might have missed it since CSV diffs are difficult to read, but it seems like we've added the columns but they're all empty, we probably want to add SMS in at least one of the tests to ensure it gets imported and exported correctly

home/models.py
fritzbrand: Cool, understood and agreed

home/models.py
fritzbrand: Yup, thx

home/tests/test_content_import_export.py
fritzbrand: Weird, I don't see them on my end, will try figure out whats up

rudigiesler (approved):

PR #233, opened 2024-01-22
Removed old importer

fritzbrand:
Removing the old importer to make the import/export code more manageable


home/tests/test_content_import_export.py
```diff
@@ -586,7 +545,7 @@ def csvs2dicts(self, src_bytes: bytes, dst_bytes: bytes) -> ExpDictsPair:
         return filter_exports(src, dst, self.importer)
 
 
-@pytest.fixture(params=["old", "new"])
+@pytest.fixture(params=["new"])
```
jerith, line 548: This doesn't need to be parameterised anymore.

home/tests/test_content_import_export.py
```diff
@@ -926,7 +861,7 @@ def test_missing_related_pages(self, newcsv_impexp: ImportExport) -> None:
 
 
 # "old-xlsx" has at least three bugs, so we don't bother testing it.
-@pytest.fixture(params=["old-csv", "new-csv", "new-xlsx"])
+@pytest.fixture(params=["new-csv", "new-xlsx"])
```
jerith, line 864: No need for the `new-` prefix here, because we now only have one importer.

jerith (commented):
May as well remove the old exporter as well. We're not even testing that, since there's no easy way to use it.

jerith (approved):

home/content_import_export.py
rudigiesler, line 1: I see there are still some classes at the end of this file that seem to be for the old importer, that I wouldn't expect us to still be using. Can those also be removed? Or are they being used somewhere that I've forgotten?

home/migrations/0044_merge_20240118_1014.py
rudigiesler, line 1: I'm not sure why we're deleting a migration here? I wouldn't expect this PR to touch any migrations

home/tests/test_content_import_export.py
```diff
@@ -110,20 +104,12 @@ def strip_leading_whitespace(entry: ExpDict) -> ExpDict:
     strip_leading_whitespace,
 ]
 
-OLD_EXPORT_FILTER_FUNCS = [
-    remove_translation_tag_from_tags,
```
jerith, line 114: I think we can remove this function (somewhere above this line) now that we don't call it anymore?

home/tests/test_content_import_export.py
```diff
@@ -444,32 +418,12 @@ def enable_web(page: DbDict) -> DbDict:
...
-OLD_PAGE_FILTER_FUNCS = [
-    remove_translation_key,
-    remove_revisions,
-    add_body_fields,
-    remove_next_prompt,
-    remove_buttons,
-    remove_example_values,
-    enable_web,
-    remove_sms_fields,
-    remove_ussd_fields,
```
jerith, line 456: We can probably remove all the functions referenced in this list.

home/tests/test_content_import_export.py
```diff
@@ -1021,15 +937,9 @@ def test_cpi_validation_failure_update(self, newcsv_impexp: ImportExport) -> Non
 
 
 # "old-xlsx" has at least three bugs, so we don't bother testing it.
```
jerith, line 939: This comment can go now that the old exporter's gone.

jerith (approved):

PR #239, opened 2024-01-29
fixed web preview

Mudiwa66:
## Purpose
On our CMS, users are not able to preview their changes. A basic template needs to be created, that will display those changes in real time.
## Approach
The user is now able to preview their changes in real time. The user can also open a new tab that will display the content on an entire webpage.

### DIFF
--- a/home/templates/home/content_page.html
+++ b/home/templates/home/content_page.html
@@ -1,0 +1,16 @@
+{% extends "base.html" %}
+{% load wagtailcore_tags %}
+
+{% block body_class %}template-contentpage{% endblock %}
+
+
+{% block content %}
+    <h1>{{ page.title }}</h1>
+    <p class="meta">{{ page.date }}</p>
+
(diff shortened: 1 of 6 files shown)


home/templates/home/content_page.html
```diff
@@ -0,0 +1,16 @@
...
+{% load wagtailcore_tags %}
+
+{% block body_class %}template-contentpage{% endblock %}
+
+
+{% block content %}
+    <h1>{{ page.title }}</h1>
+    <p class="meta">{{ page.date }}</p>
+
+    <div class="intro">{{ page.intro }}</div>
```
rudigiesler, line 11: I don't think we have `date` and `intro` fields in our `ContentPage` model. We do have `title`, `subtitle`, and `body` though (at least for the web section of the page)

rudigiesler (commented):
Just one question about the fields that have been chosen to display

rudigiesler (approved):

PR #243, opened 2024-02-01
Import test

Mudiwa66:
## Purpose
We do not have a test that ensures that optional import fields are and will be truly optional

## Approach
There is a test added that imports a CSV with only those required fields populated.

When importing a csv with only the required fields present(ie. the columns of the csv), the UI import works. However, when testing that csv, and comparing it with the exported file, the test fails. This is because the exporter will generate all the columns for that file, while the imported csv will only have a few columns. The fix to this is to  test using a csv that has all the fields, but only the required fields are populated.

### DIFF
--- a/home/tests/required_fields.csv
+++ b/home/tests/required_fields.csv
@@ -1,0 +1,2 @@
+message,slug,parent,web_title,locale
+0,main-menu,,Main Menu,English
--- a/home/models.py
+++ b/home/models.py
@@ -7,7 +7,6 @@ from django.db import models
 from django.db.models.signals import pre_save
 from django.dispatch import receiver
 from django.forms import CheckboxSelectMultiple
-from django.utils.text import slugify
 from django.utils.translation import gettext_lazy as _
 from modelcluster.contrib.taggit import ClusterTaggableManager
 from modelcluster.fields import ParentalKey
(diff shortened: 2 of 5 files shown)


home/tests/required_fields.csv
```diff
@@ -0,0 +1,3 @@
+structure,message,page_id,slug,parent,web_title,locale
```
rudigiesler, line 1: Are structure and page_id required for the import? What happens if we remove those?

rudigiesler (commented):
Looks good! I think we do want the `required_fields.csv` format, since that's what we actually want to test, if the user leaves out those fields from the import file, does it still work.

In terms of the test, yes if we use the existing testing tools it will fail because of those fields that get generated and set to their defaults by the exporter. We'll need to remove those optional fields in the test before comparing them

home/tests/required_fields.csv
Mudiwa66: The import on the UI still works s we can remove those as well

jerith (commented):
We seem to have lost some tests. Can you put them back, please? 🙂

jerith:
Also, we have two different kinds of pages with different requirements. The test you've added here is for a `ContentPageIndex`, so we'll need at least one more for a `ContentPage`.

Mudiwa66:
This branch got too messy so I moved to a fresh one. Closed.

PR #246, opened 2024-02-05
reworked import test+ worked on fresh branch

Mudiwa66:
Fixed the import test. Check the db state as opposed to comparing the input and output files directly. Also switched to new branch since the old one got too messy.

### DIFF
--- a/home/tests/test_content_import_export.py
+++ b/home/tests/test_content_import_export.py
@@ -862,7 +862,7 @@ class TestImportExport:
         for page_data, content_page in zip(src, content_pages, strict=False):
             assert page_data.get("slug", "") == content_page.slug
 
-        assert content_pages.count() == len(src)
+        assert content_pages.count() == len(list(src))
 
     def test_ContentPage_required_fields(self, csv_impexp: ImportExport) -> None:
         """
@@ -881,7 +881,7 @@ class TestImportExport:
         for page_data, content_page in zip(src, content_pages, strict=False):
             assert page_data.get("slug", "") == content_page.slug
 
-        assert len(content_pages) == len(src)
+        assert len(content_pages) == len(list(src))
 
 
 @pytest.fixture(params=["csv", "xlsx"])
--- a/home/tests/test_content_import_export.py
+++ b/home/tests/test_content_import_export.py
@@ -847,6 +847,42 @@ class TestImportExport:
             == "Validation error: {'translation_key': ['“BADUUID” is not a valid UUID.']}"
         )
 
+    def test_ContentPageIndex_required_fields(self, csv_impexp: ImportExport) -> None:
+        """
+        Importing an CSV file with only the required fields shoud not break
+
+        """
+
+        csv_bytes = csv_impexp.import_file("contentpage_index_required_fields.csv")
+        content = csv_impexp.export_content()
+        src, dst = csv_impexp.csvs2dicts(csv_bytes, content)
+
+        content_pages = ContentPageIndex.objects.all()
+
+        for page_data, content_page in zip(src, content_pages, strict=False):
+            assert page_data.get("slug", "") == content_page.slug
+
+        assert content_pages.count() == len(src)
+
+    def test_ContentPage_required_fields(self, csv_impexp: ImportExport) -> None:
+        """
+        Importing an CSV file with only the required fields shoud not break
+
+        """
+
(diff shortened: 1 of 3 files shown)


Mudiwa66:
This took longer than anticipated.

home/tests/test_content_import_export.py
```diff
@@ -847,6 +847,42 @@ def test_cpi_validation_failure_update(self, csv_impexp: ImportExport) -> None:
             == "Validation error: {'translation_key': ['“BADUUID” is not a valid UUID.']}"
         )
 
+    def test_ContentPageIndex_required_fields(self, csv_impexp: ImportExport) -> None:
+        """
+        Importing an CSV file with only the required fields shoud not break
+
```
jerith, line 853: No need for these empty lines at the end of the docstrings.

home/tests/test_content_import_export.py
```diff
@@ -847,6 +847,42 @@ def test_cpi_validation_failure_update(self, csv_impexp: ImportExport) -> None:
...
+        """
+
+        csv_bytes = csv_impexp.import_file("contentpage_index_required_fields.csv")
+        content = csv_impexp.export_content()
+        src, dst = csv_impexp.csvs2dicts(csv_bytes, content)
+
+        content_pages = ContentPageIndex.objects.all()
```
jerith, line 860: Since we know what the source data is, we can be a lot more explicit with the assertions:

```python
[main_menu] = ContentPageIndex.objects.all()
assert main_menu.slug == "main-menu"
```

The first line will fail if we don't have exactly one `ContentPageIndex`, and the second will fail if the slug doesn't match.

home/tests/test_content_import_export.py
```diff
@@ -847,6 +847,42 @@ def test_cpi_validation_failure_update(self, csv_impexp: ImportExport) -> None:
...
+
+        """
+
+        csv_bytes = csv_impexp.import_file("contentpage_required_fields.csv")
+        content = csv_impexp.export_content()
+        src, dst = csv_impexp.csvs2dicts(csv_bytes, content)
+
+        content_pages = list(ContentPageIndex.objects.all()) + list(
+            ContentPage.objects.all()
+        )
```
jerith, line 879: As with the previous test, we can be a lot more explicit:

```python
[main_menu] = ContentPageIndex.objects.all()
[first_time_user, health_info, self_help] = ContentPage.objects.all()
# And then a slug assertion for each page.
```

Mudiwa66: Got it, I kept the line that defines `content_pages` so checking the sum a a bit cleaner.

home/tests/test_content_import_export.py
Mudiwa66: I kept this line since I use that to check how many pages/ indexes we have, but I could find a work around to get that data directly from `csv_bytes`, not sure.

jerith: Comparing the original and re-exported CSVs will ensure that the line/page counts are the same, and the `[main_menu] = ...` assignment will raise an exception if the sequence on the right doesn't have exactly one element. 🙂

home/tests/test_content_import_export.py
jerith: Do we really need to assert on the sum if we already get an exception when any of the queries returns the wrong number of results?

Mudiwa66: Correct, no need for that check, removed.

home/tests/test_content_import_export.py
```diff
@@ -847,6 +847,48 @@ def test_cpi_validation_failure_update(self, csv_impexp: ImportExport) -> None:
...
+    def test_ContentPageIndex_required_fields(self, csv_impexp: ImportExport) -> None:
+        """
+        Importing an CSV file with only the required fields for a ContentPageIndex shoud not break
+        """
+
+        csv_bytes = csv_impexp.import_file("contentpage_index_required_fields.csv")
+        content = csv_impexp.export_content()
+        src, dst = csv_impexp.csvs2dicts(csv_bytes, content)
+
+        # the importer adds extra fields, so we fikter for the ones we want
```
jerith, line 859: Tyop in this comment (and the one below): `fikter` -> `filter`.

jerith (approved):

PR #249, opened 2024-02-08
changed import name and fixed typo in test

Mudiwa66:
## Purpose
The import button on the edit page is to import a word document into the web section of the page, so we want to change to text to "Import web from doc"

## Approach
The import button is more descriptive


Mudiwa66:
@jerith any idea why this would fail? It works on my machine, looks like some missing css content. All my migrations are up to date as well.

jerith:
As discussed on slack: We're apparently using a staticfiles storage backend that isn't test-friendly. The quickest fix is probably to change the storage backend in the dev or test settings file.

Mudiwa66:
Seems to be working now.

jerith (approved):
🎉

PR #251, opened 2024-02-13
Fix crash when exporting data with media or images.

HawkiesZA:
## Purpose
To allow exporting to xlsx when pages have media. (Link to ticket.)[https://praekelt.leankit.com/card/31512190149211]

## Approach
Instead of exporting the entire image / media, this exports the URL


rudigiesler (dismissed):
Looks good!

rudigiesler (approved):

PR #252, opened 2024-02-13
Update release process

rudigiesler:
## Purpose
The current release procedure is complex and difficult to understand. It also doesn't match our current development speed, and users often want features much earlier than releases.

## Approach
This PR puts together a proposal for a simpler release process, that hopefully better matches the development and usage of contentrepo

It also adds documentation on how to actually create a release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,26 +28,29 @@ This content repository allows easy content management via a headless CMS, using
 
 
 ## Releases
-The current LTS release is: 1.0
-
 [Semantic versioning](https://semver.org/) is in use in this project.
-1. Feature releases
-    - Feature releases are a MAJOR.MINOR combination, eg. `1.2`
-    - Feature releases occur once every 3 months, but only if required. eg. If there were only patch releases in the last 3 months, no feature release will occur.
-    - The latest feature release will receive any security or bug fixes as patch releases
-    - Each feature release will have a separate git branch
-1. Patch releases
-    - Patch releases are used to fix bugs and security issues
-    - They are released as soon as they are ready
-1. LTS (Long-Term Support) Releases
-    - Every fourth release is an LTS release
-    - LTS releases are listed in the documentation
-    - LTS releases receive security patches
-    - LTS releases are supported for 5 feature releases, allowing for 15 months of support with a 3 month switchover time to the next LTS
+
 1. Development releases
-    - Development have the `-dev.N` suffix
-    - They are used to test new code before an official release is made
+    - Development versions have the `-dev.N` suffix
+    - They are used to test new code in QA
     - They are built off of the `main` git branch
+    - They are created with git tags, eg. `v1.2.0-dev.2`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,26 +28,29 @@ This content repository allows easy content management via a headless CMS, using
...
-    - Every fourth release is an LTS release
-    - LTS releases are listed in the documentation
-    - LTS releases receive security patches
-    - LTS releases are supported for 5 feature releases, allowing for 15 months of support with a 3 month switchover time to the next LTS
+
 1. Development releases
-    - Development have the `-dev.N` suffix
-    - They are used to test new code before an official release is made
+    - Development versions have the `-dev.N` suffix
+    - They are used to test new code in QA
```
jerith, line 35: Do we want a new dev release for each feature that goes to QA?

It might make sense once the pace of development slows down somewhat, but at the moment I expect it'll be more of a barrier than a help.

README.md
```diff
@@ -28,26 +28,29 @@ This content repository allows easy content management via a headless CMS, using
...
+    - Releases are created when the changes have passed QA
     - They are built off of the `main` git branch
+    - They are created with git tags, eg. `v1.2.0`
+
+### How to create a release
+
+1. Check that `CHANGELOG.md` is up to date.
+1. Update the version number in `pyproject.toml` to the version you want to release, eg. change `1.2.0-dev.0` to `1.2.0`
+1. Either create a pull request with these changes, and have it reviewed, or post a diff in Slack and have it reviewed there. Then merge, or commit and push, the changes.
+1. Tag these changes with the desired release, eg. `git tag v1.2.0`, and push the tag, eg. `git push origin v1.2.0`
```
jerith, line 50: `git push --tags` is probably a better option than `git push origin v1.2.0`.

README.md
rudigiesler: Yeah I guess we don't need it, we can rely on just the SHA hashes. Should we get rid of dev releases completely then?

README.md
rudigiesler: Won't that push all tags? I guess you don't usually make tags except when doing releases

README.md
jerith: Dev releases may still be useful, especially if we have some experimental features that we want to deploy somewhere less volatile than a shared QA environment. I think we can leave them in, but remove the item about using them for QA deployments.

Somewhat unrelated: Do we want a version check in version-tagged image builds to make sure the `pyproject.toml` version matches the tag?

README.md
jerith: It will, but I can't think of any reason to have local tags that shouldn't be pushed.  🤷

README.md
rudigiesler: Yeah we can add that, good idea. You can move a tag to a different commit right?

fritzbrand (approved):
Thanks, looks good to me

jerith (approved):

PR #259, opened 2024-02-19
Ordered set export

HawkiesZA:
## Purpose
The ordered content set export currently only includes the list of pages and not the time, unit, before or after, or contact fields

## Approach
This includes all the required fields in the export.


home/wagtail_hooks.py
```diff
@@ -200,6 +207,48 @@ def page(self, obj):
 
     page.short_description = "Page Slugs"
 
+    def time(self, obj):
+        if obj.pages:
+            return [
+                (
+                    f"{p.value['time']} {p.value['unit']}"
+                    if p.value and "time" in p.value
```
rudigiesler, line 215: Should we maybe have unit be a separate field, or add it to this if statement? Otherwise it is possible to have time but no unit, since they're both optional fields.

rudigiesler (commented):
Looks good, just one comment about combining the `time` and `unit` fields.

Then in the `content_import_export`, there's a `import_ordered_sets` import function, where we should probably also add these fields to the import, otherwise we can export but not import them

home/wagtail_hooks.py
HawkiesZA: Good point. I've made it a separate field.

rudigiesler (approved):

PR #263, opened 2024-02-22
Improve TestContentPageAPI tests

erikh360:
## Purpose
Improving some tests on TestContentPageAPI, mainly splitting them out making sure each test only tests one specific thing.

## Approach
fast but careful


jerith (approved):
We should probably create each test's pages in the test rather than rummaging through a shared set, but that can be a separate PR.

PR #264, opened 2024-02-22
Omega 51 contentrepo export broken in prod instance

Buhle79:
## Purpose

## Approach

## Learning
_Describe the research stage_

#### Blog Posts


home/export_content_pages.py
```diff
@@ -147,10 +147,15 @@ def add_message_fields(self, msg_blocks: MsgBlocks) -> None:
     @staticmethod
     def serialise_buttons(buttons: blocks.StreamValue.StreamChild) -> str:
         button_dicts = []
+
         for button in buttons:
             button_dict = {"type": button.block_type, "title": button.value["title"]}
             if button.block_type == "go_to_page":
+                # Exclude buttons that has deleted pages that they are linked to it
+                if button.value.get("page") is None:
+                    continue
```
jerith, line 156: Did we confirm that this is the desired behaviour for now, rather than failing the export with an error pointing to the problem?

Buhle79: Yes for now, this is something we need to handle in the exporter. Handle it the same way we handle Related pages in the exporter. @jerith

home/tests/test_content_import_export.py
```diff
@@ -1882,3 +1884,128 @@ def test_list_items(self, impexp: ImportExport) -> None:
         impexp.export_reimport()
         imported = impexp.get_page_json()
         assert imported == orig
+
+    def test_content_page_with_no_go_to_button(self, impexp: ImportExport) -> None:
+        """
+        If a page has a button go to page it should export all buttons.
```
jerith, line 1890: The test name and docstring don't seem to agree, and I'm not really sure what this is intended to test. Can you clarify the docstring and make sure the test name matches, please?

home/tests/test_content_import_export.py
```diff
@@ -1882,3 +1884,128 @@ def test_list_items(self, impexp: ImportExport) -> None:
...
+                        )
+                    ],
+                )
+            ],
+        )
+
+        orig = impexp.get_page_json()
+
+        orig_btn = (
+            # list(orig)[3].get("fields").get("whatsapp_body")[0].get("value").get("buttons")
```
jerith, line 1995: If we don't need this commented code anymore, let's delete it rather.

home/export_content_pages.py
rudigiesler: @jerith Yes I think this is the behaviour we want. Failing the import with an error message works because the user could give us any malformed file, and we want to tell them how to fix that file. But the export is from the database, which should only have valid data in it, so any issues with that I would consider a bug, and not something for the user to fix.

In this case we probably want to fix having null pages, but I don't know how easy a cascading delete will be here since it's not a normal table relation, but for now the short-term fix is to skip any buttons without pages (similar to if we had to cascade on delete)

home/tests/test_content_import_export.py
```diff
@@ -1882,3 +1884,128 @@ def test_list_items(self, impexp: ImportExport) -> None:
...
+        impexp.export_reimport()
+        imported = impexp.get_page_json()
+        imported_btn = list(imported)[3]["fields"]["whatsapp_body"][0]["value"][
+            "buttons"
+        ]
+
+        assert imported != orig
+        assert imported_btn != orig_btn
+        assert imported_btn == []
+        assert orig_btn != []
```
jerith, line 2011: In general, asserts should assert on what things *are*, not what they *aren't*.

If we expect the original and reimported pages to be different, we should explicitly modify the original page json to match what we expect the reimported version to look like and assert that it does, otherwise we may miss later bugs that cause the export to change in other ways.

home/tests/test_content_import_export.py
```diff
@@ -1896,3 +1898,125 @@ def test_list_items(self, impexp: ImportExport) -> None:
...
+        impexp.export_reimport()
+        imported = impexp.get_page_json()
+        imported_btn = list(imported)[3]["fields"]["whatsapp_body"][0]["value"][
+            "buttons"
+        ]
+
+        assert imported_btn == []
+        assert orig_btn == [
+            {"type": "go_to_page", "value": {"page": 1, "title": "Import Export"}}
+        ]
```
jerith, line 2022: We need to assert on the rest of the content as well, otherwise there could be other breakages that we don't pick up in the test.

home/tests/test_content_import_export.py
```diff
@@ -1896,3 +1909,178 @@ def test_list_items(self, impexp: ImportExport) -> None:
...
+
+    def test_export_with_missing_button(self, impexp: ImportExport) -> None:
+        """
+        If parent page that is linked to a button gets deleted we skip that button.
+        """
```
jerith, line 2038: This tests the same thing as `test_export_missing_go_to_button`, but in a better way. The other test has a better name and docstring, though. Can you combine them into one test with name and doctring from `test_export_missing_go_to_button` and the code from this one, please?

jerith (approved):
🎉

PR #268, opened 2024-02-26
Check that the value is not none to stop the export from crashing

HawkiesZA:
## Purpose
Sometimes the export crashes because the values in the list being exported are None.

## Approach
This checks if the values are None, and makes it an empty string if it is None.


home/wagtail_hooks.py
```diff
@@ -211,7 +211,11 @@ def page(self, obj):
     def time(self, obj):
         if obj.pages:
             return [
-                (f"{p.value['time']}" if p.value and "time" in p.value else "")
+                (
+                    f"{p.value['time']}"
+                    if p.value and "time" in p.value and p.value["time"]
```
rudigiesler, line 216: Is `p.value` here just a dictionary? Could we do `f"{p.value["time"]}" if p.value and p.value.get("time") else ""`?

Or maybe we can split this out into a method, since this is repeated, and rather do the ask for forgiveness pattern?
```
def get_field_value(page: Page, field: str) -> str:
    try:
        if value := p.value[field]:
            return f"{value}"
        else:
            return ""
    except AttributeError, TypeError:
        return ""
```

rudigiesler (commented):
Just a comment on how we might better structure this, since it seems to be growing, but otherwise looks good, I think this will work. Is it possible to add a test for this, or are we not sure how the data got into this shape in the first place, and so it's difficult to write a test for?

rudigiesler (dismissed):

rudigiesler (approved):

PR #270, opened 2024-02-28
Update TestWhatsAppMessages

erikh360:


home/tests/test_api.py
```diff
@@ -462,93 +464,95 @@ def test_detail_view_no_content_page(self, uclient):
 @pytest.mark.django_db
 class TestWhatsAppMessages:
     """
-    FIXME:
-     * Should some of the WhatsApp tests from TestPagination live here instead?
```
erikh360, line 466: I kept these separate because they test the WhatsApp specific functionality where the WhatsApp tests in TestPagination tests the more generic things but just using the WhatsApp channel. 

Let me know if you think otherwise.

home/tests/test_api.py
```diff
@@ -462,93 +464,95 @@ def test_detail_view_no_content_page(self, uclient):
...
-        page.save_revision().publish()
+        page = self.create_content_page(buttons=[NextBtn("Tell me more")])
 
         response = uclient.get(f"/api/v2/pages/{page.id}/?whatsapp=true&message=1")
         content = response.json()
         [button] = content["body"]["text"]["value"]["buttons"]
         button.pop("id")
         assert button == {"type": "next_message", "value": {"title": "Tell me more"}}
 
-    def test_whatsapp_template(self, uclient):
```
erikh360, line 499: I merged this one into `test_whatsapp_template_fields`, better to just check all the template fields in one go.

home/tests/test_api.py
```diff
@@ -462,93 +464,95 @@ def test_detail_view_no_content_page(self, uclient):
...
-        """
-        FIXME:
-         * Is this actually a template message?
```
erikh360, line 502: I'm not sure if it was or not with the way it was created here, but the PageBuilder does set `is_whatsapp_template` to True if you specify a template name or category.

home/tests/test_api.py
jerith: Yeah, I think that was the intention and I just expressed it poorly in the FIXME.

home/tests/test_api.py
jerith: That FIXME may have been related to a previous version of the test. 🤷

PR #273, opened 2024-02-28
Add tests for functions in serializers

megh-k:
## Purpose
Adding more test coverage to the code base. 

## Approach
It does not seem straight-forward to test the serializers, wagtail's codebase does not directly test them and instead, tests the data in the API. We will take this approach for the serializer but we have 3 functions that still need to be tested, `has_next_message`, `has_previous_message` and `format_whatsapp_message`. Some logic was added where tests failed

## TODO

- [ ] Look through tests and add more where logic is untested

### DIFF
--- a/home/tests/test_serializers.py
+++ b/home/tests/test_serializers.py
@@ -69,10 +69,6 @@ def create_content_page(
         slug=title.replace(" ", "-"),
         title=title,
         bodies=bodies,
-        tags=[],
-        quick_replies=[],
-        triggers=[],
-        publish=True,
     )
     return content_page
 


home/tests/test_serializers.py
```diff
@@ -0,0 +1,388 @@
...
+    # Test has_next_message
+    def test_has_next_message_whatsapp(self) -> None:
+        """Checks if the next message matches up.
+        Index of first message is 0, next message is then message 2 with index 1"""
+        page = self.create_content_page(wa_body_count=2)
+        assert page.whatsapp_body
+        assert has_next_message(0, page, "whatsapp") == 2
+
+    def test_has_next_message_viber_no_next_message(self) -> None:
+        """On a page with 2 viber messages, the next message is None"""
```
jerith, line 94: Do we have two viber messages here? Also, is the `None` because we're looking at viber messages, or because we pass index 1 to `has_next_message`?

home/tests/test_serializers.py
```diff
@@ -0,0 +1,388 @@
...
+        page = self.create_content_page()
+        assert page.viber_body
+        assert has_next_message(1, page, "viber") is None
+
+    def test_has_next_message_invalid_platform(self) -> None:
```
jerith, line 99: We don't need `has_next_message` in the test name, because that's already in the class name. Perhaps `test_no_next_message_for_invalid_platform` would be a better name here?

(Similarly for various other test names.)

home/tests/test_serializers.py
```diff
@@ -0,0 +1,388 @@
...
+@pytest.mark.django_db
+class TestHasNextMessage:
+    def create_content_page(
+        self,
+        parent: Union[ContentPage, None] = None,
+        title: str = "default page",
+        tags: Union[List[str], None] = None,
+        wa_body_count: int = 1,
+        publish: bool = True,
+    ) -> ContentPage:
```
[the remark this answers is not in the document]
megh-k: The last one is different as it adds variations, but I can just use that version!

home/tests/test_serializers.py
megh-k: Oops, "on a page with 1 viber message"

jerith: This docstring is still confusing, as is the one above which is identical despite having a different message index.

As I understand it, `has_next_message()` should return the index of the message after the given index (for the given platform) if there is one, otherwise `None`. If so, perhaps these might be better docstrings given the code for these two tests:

"The next message after the last message is None."
"The next message after a message that doesn't exist is None."

However, one of the test names indicates that we should get a non-None result? 🤷

home/tests/test_serializers.py
```diff
@@ -0,0 +1,258 @@
...
+        quick_replies=[],
+        triggers=[],
+        publish=True,
+    )
+    return content_page
+
+
+@pytest.mark.django_db
+class TestHasNextMessage:
+    # Test has_next_message
```
jerith, line 105: This comment is redundant with the class name. If you feel the name is insufficient, rather add a docstring with more information.

home/tests/test_serializers.py
```diff
@@ -0,0 +1,258 @@
...
+
+    def test_no_next_message_empty_body(self) -> None:
+        """No next message on a page with no whatsapp body"""
+        page = create_content_page(wa_body_count=0)
+        assert has_next_message(0, page, "whatsapp") is None
+
+    def test_no_next_message_on_last_message(self) -> None:
+        """last messages' next message is None"""
+        page = create_content_page(wa_body_count=1)
+        assert has_next_message(1, page, "whatsapp") is None
```
jerith, line 143: The name and docstring say this test is about the last message, but with only one message the last index will be 0. Perhaps create two messages here?

home/tests/test_serializers.py
```diff
@@ -0,0 +1,258 @@
...
+
+    def test_next_message_on_first_message_of_many(self) -> None:
+        """Page with many whatsapp messages has a next message on the first"""
+        page = create_content_page(wa_body_count=5)
+        assert has_next_message(0, page, "whatsapp") == 2
```
jerith, line 148: Mixing zero-indexed (for the input) and one-indexed (for the output) values creates all kinds of confusion. We should probably fix the code under test to be consistent. (That's a job for another PR, though.)

home/tests/test_serializers.py
megh-k: damn I just realised that I forgot that the message number != index, thursday meg seems to have been very very unfocused, apologies, fixing these two tests

home/tests/test_serializers.py
```diff
@@ -0,0 +1,258 @@
...
+    bodies = [
+        MBody(
+            title,
+            [
+                MBlk("*Default Messenger Content* 🏥1"),
+                MBlk("*Default Messenger Content* 2🏥"),
+            ],
+        ),
+        SBody(
+            title, [SBlk("*Default SMS Content* 1"), SBlk("*Default SMS Content*2 ")]
```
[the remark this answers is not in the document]
megh-k: Unfortunately changing the bodies to `{platform name} {message number}` is still a bit long, we could shorten it down to just `{message number}` buuuut then there would be no differentiation between platforms, so I think we are going to have to just accept the unfortunately multiline situation

home/tests/test_serializers.py
jerith: This is why mixing zero-indexed and one-indexed values is such a terrible idea. It basically guarantees this sort of mistake. 🤷

home/tests/test_serializers.py
```diff
@@ -102,25 +79,60 @@ def create_content_page(
...
+        """SMS The next message after the last message is None."""
+        page = create_content_page()
+        assert page.sms_body
+        assert has_next_message(2, page, "sms") is None
+
+    def test_next_message_for_sms(self) -> None:
+        """SMS Next message after 1st message (index 0) is message 2 (index 1)"""
+        page = create_content_page()
+        assert page.sms_body
+        assert has_next_message(0, page, "sms") == 2
```
jerith, line 111: We can probably combine these into a single test per platform:
```
        page = create_content_page()
        assert page.sms_body
        assert has_next_message(0, page, "sms") == 2
        assert has_next_message(2, page, "sms") is None
```

megh-k:
@jerith I added some more tests and more assertions in the most recent commit, decided its probably best to check that what we expect to be there is there and that the functions work for all the platforms

home/tests/test_serializers.py
```diff
@@ -0,0 +1,285 @@
...
+        UBody(title, [UBlk("USSD 1"), UBlk("USSD 2")]),
+        VBody(title, [VBlk("Viber 1"), VBlk("Viber 2")]),
+    ]
+
+    for i in range(wa_body_count):
+        wa_body = f"WhatsApp {i+1}"
+
+        variation_messages = [
+            VarMsg(f"WhatsApp Varied {var+1}", gender=wa_gender_var[var])
+            for var in range(len(wa_gender_var))
```
jerith, line 60: `[VarMsg(f"WhatsApp Varied {i}", gender=gender) for i, gender in enumerate(wa_gender_var, 1)]` perhaps?

home/tests/test_serializers.py
```diff
@@ -0,0 +1,285 @@
...
+    content_page = PageBuilder.build_cp(
+        parent=parent,
+        slug=title.replace(" ", "-"),
+        title=title,
+        bodies=bodies,
+        tags=[],
+        quick_replies=[],
+        triggers=[],
+        publish=True,
+    )
```
jerith, line 76: The last four fields here have defaults, so we don't need to pass them in.

home/tests/test_serializers.py
```diff
@@ -0,0 +1,285 @@
...
+    return content_page
+
+
+@pytest.mark.django_db
+class TestHasNextMessage:
+    def test_next_messages_whatsapp(self) -> None:
+        """WhatsApp The next message after the last message is None.
+        Index of first message is 0, next message is then message 2 with index 1"""
```
jerith, line 84: I don't think we've explicitly standardised on docstring formatting, but in most places we put the `"""` on their own line. Can we do the same in these tests?

home/tests/test_serializers.py
```diff
@@ -0,0 +1,285 @@
...
+    def test_with_invalid_index(self) -> None:
+        """Page with 1 wa message should throw an error on request for 11th message"""
+        page = create_content_page(wa_body_count=1)
+        assert page.whatsapp_body
+        assert len(page.whatsapp_body) == 1
+        with pytest.raises(IndexError):
+            format_whatsapp_message(10, page, "whatsapp")
+
+    def test_happy_case_single_variation_message(self) -> None:
+        page = create_content_page(wa_body_count=1, wa_gender_var=["female"])
```
jerith, line 238: Missing docstring?

home/tests/test_serializers.py
```diff
@@ -0,0 +1,285 @@
...
+                "list_items": [],
+                "footer": "",
+            },
+            "id": f'{result["id"]}',
+        }
+
+        assert result == expected_result
+
+    def test_happy_case_no_variation_messages(self) -> None:
+        page = create_content_page(wa_body_count=1)
```
jerith, line 267: Missing docstring?

PR #276, opened 2024-02-29
Upgrade wagtail to 5.1.3 to get ordered content sets import and export working

HawkiesZA:
## Purpose
Moving ordered content sets to snippets accidentally removed the import and export buttons.

## Approach
Upgrading wagtail to 5.1.3 gives us the export functionality for free with the upgrade. From there I modified the template to get the import button back similar to how we used to do it with ModelAdmin. In a later upgrade of Wagtail, the template can be simplified more.

- [x] In orderedcontentset_upload.html the `destinationURL` has now been set to a static var because I can't seem to figure out how the dynamic referencing works (if it does even). Is that ok?

## Learning
[Wagtail 5.1 upgrade release notes for snippets](https://docs.wagtail.org/en/stable/releases/5.1.html#snippet-enhancements)
[Future niceness for the template](https://github.com/wagtail/wagtail/pull/11278/files)


rudigiesler (approved):
Not sure if there's anything we can do to not have to duplicate so much of the base template, but looks good to me

HawkiesZA:
> Not sure if there's anything we can do to not have to duplicate so much of the base template, but looks good to me

I looked into it, but it looks like not at this time. Once we've updated to Wagtail 5.2 we should be able to.

PR #282, opened 2024-03-07
Switch to poetry

jerith:
## Purpose
Clean up dependency management and get rid of multiple conflicting requirements files.

## Approach
Switching to poetry, which we use for most of our other (newer) Python projects, gives us a single source of truth and better tooling for dependency management. Since our dependencies were unclear anyway, this also updates to the latest LTS versions of Django and Wagtail (which we were on or near already) and the latest versions of everything else.


pyproject.toml
```diff
@@ -1,28 +1,84 @@
...
+requires = ["poetry-core"]
+build-backend = "poetry.core.masonry.api"
 
-[project]
+[tool.poetry]
 name = "contentrepo"
 version = "1.1.1-dev.0"
-authors = [
-  { name="Praekelt.org", email="dev@praekelt.org" },
-]
```
erikh360, line 10: When do we change this to reach?

erikh360 (approved):

pyproject.toml
jerith: Open a PR? 🙂

erikh360 (approved):

PR #283, opened 2024-03-07
Test api

megh-k:
## Purpose
Increase test coverage for API and serializers

## Approach
add the following tests

- test_messenger_disabled
- test_web_disabled
- test_viber_disabled
- test_sms_disabled
- test_ussd_disabled
- test_wa_image
- test_viber_image
- test_messenger_image
- test_wa_doc
- test_wa_media
- test_wa_messages_only_1_image
- test_wa_messages_only_1_media
- test_wa_messages_only_1_doc
- test_whatsapp_template_variables
- test_list_items
- test_next_prompt
- test_footer
- test_empty_whatsapp

Note on image/doc/media testing: 
All messages that are created with the image/doc/media parameter will allocate the same image. This is done to minimise complexity. A test is added to ensure that only one image is created when multiple messages are created 

Fix: 
Incorrect Type in import tests

### DIFF
--- a/home/tests/test_api.py
+++ b/home/tests/test_api.py
@@ -4,7 +4,12 @@ from pathlib import Path
 
 import pytest
 from bs4 import BeautifulSoup
+from django.core.files.base import File  # type: ignore
+from django.core.files.images import ImageFile  # type: ignore
 from pytest_django.asserts import assertTemplateUsed
+from wagtail.documents.models import Document  # type: ignore
+from wagtail.images.models import Image  # type: ignore
+from wagtailmedia.models import Media  # type: ignore
 
 from home.content_import_export import import_content
 from home.models import (
@@ -24,15 +29,19 @@ from .page_builder import (
     UBlk,
     UBody,
     VarMsg,
+    VBlk,
+    VBody,
     WABlk,
     WABody,
 )
 from .utils import create_page
 
 
-# use this to access the admin interface
 @pytest.fixture()
 def admin_client(client, django_user_model):
+    """
+    Access admin interface
+    """
     creds = {"username": "test", "password": "test"}
     django_user_model.objects.create_superuser(**creds)
     client.login(**creds)
@@ -41,10 +50,51 @@ def admin_client(client, django_user_model):
 
 @pytest.fixture()
 def uclient(client, django_user_model):
+    """
+    Access the user interface
+    """
     creds = {"username": "test", "password": "test"}
     django_user_model.objects.create_user(**creds)
     client.login(**creds)
     return client
+
+
+def mk_test_img() -> Image:
+    """
+    make test image
+    """
+    img_path = Path("home/tests/test_static") / "test.jpeg"
+    img = Image(
+        title="default image title",
+        file=ImageFile(img_path.open("rb"), name=img_path.name),
+    )
+    img.save()
+    return img
+
+
+def mk_test_media() -> File:
+    """
+    make test media
+    """
+    media_path = Path("home/tests/test_static") / "test.txt"
+    media = Media(
+        title="default media title",
+        file=File(media_path.open("rb"), name=media_path.name),
+    )
+    media.save()
+    return media
+
+
+def mk_test_doc() -> Document:
+    """
+    make test document
+    """
+    doc_path = Path("home/tests/test_static") / "test.mp4"
+    doc = Document(
+        title="default doc title", file=File(doc_path.open("rb"), name=doc_path.name)
+    )
+    doc.save()
+    return doc
 
 
 @pytest.mark.django_db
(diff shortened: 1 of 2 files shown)


home/tests/test_api.py
```diff
@@ -41,12 +50,53 @@ def admin_client(client, django_user_model):
...
     creds = {"username": "test", "password": "test"}
     django_user_model.objects.create_user(**creds)
     client.login(**creds)
     return client
 
 
+def mk_test_img() -> Image:
+    """
+    make test image
+    """
```
jerith, line 65: For functions where the purpose is adequately expressed in the name, having that repeated in a docstring doesn't really add much value. 🙂

home/tests/test_api.py
```diff
@@ -41,12 +50,53 @@ def admin_client(client, django_user_model):
...
+    img_path = Path("home/tests/test_static") / "test.jpeg"
```
jerith, line 66: Would it make sense for `Path("home/tests/test_static")` to be a module-level constant or something?

home/tests/test_api.py
```diff
@@ -41,12 +50,53 @@ def admin_client(client, django_user_model):
...
+    )
+    img.save()
+    return img
+
+
+def mk_test_media() -> File:
+    """
+    make test media
+    """
+    media_path = Path("home/tests/test_static") / "test.txt"
```
jerith, line 79: Is `test.txt` "media"?

home/tests/test_api.py
```diff
@@ -41,12 +50,53 @@ def admin_client(client, django_user_model):
...
+    )
+    media.save()
+    return media
+
+
+def mk_test_doc() -> Document:
+    """
+    make test document
+    """
+    doc_path = Path("home/tests/test_static") / "test.mp4"
```
jerith, line 92: Is `test.mp4` a "document"?

home/tests/test_api.py
```diff
@@ -94,16 +153,41 @@ def create_content_page(
...
             if body_type == "WhatsApp":
-                bodies.append(WABody(title, [WABlk(msg_body)]))
+                bodies.append(
+                    WABody(
+                        title,
+                        [
+                            WABlk(
+                                msg_body,
+                                image=image_id,
+                                media=media_id,
```
jerith, line 177: We could probably save some vertical space by extracting the `WABlk(...)` to a local variable and passing it to `WABody(...)`.

home/tests/test_api.py
```diff
@@ -94,16 +153,41 @@ def create_content_page(
...
+                                document=doc_id,
```
jerith, line 178: Do we really want all images/docs/media in every message? If not, maybe building pages by hand in the relevant tests would be cleaner.

home/tests/test_api.py
```diff
@@ -281,6 +365,56 @@ def test_whatsapp_disabled(self, uclient):
         response = uclient.get(f"/api/v2/pages/{page.id}/?whatsapp=True")
         assert response.content == b""
 
+    def test_messenger_disabled(self, uclient):
+        """
+        It should not return the web body if enable_messenger=false
```
jerith, line 370: Web body? (Also other tests below.)

home/tests/test_api.py
```diff
@@ -460,6 +594,115 @@ def test_detail_view_no_content_page(self, uclient):
...
+        response = uclient.get(f"/api/v2/pages/{page.id}/?whatsapp=true")
+        content = response.json()
+
+        doc_id = content["body"]["text"]["value"]["document"]
+        assert doc_id == page.whatsapp_body._raw_data[0]["value"]["document"]
+
+    def test_wa_messages_only_1_image(self, uclient):
+        """
+        Test that when many images are assigned to messages in the test, only 1 image is created in the test
+        """
```
jerith, line 655: Having tests for our local test helpers may be an indication that they're too complex and we should rather be more explicit in the tests themselves.

jerith (commented):
There are a lot of tests that are identical except for the message platform we're looking at. We can probably use pytest's parameterisation features to combine and simplify those.

home/tests/test_api.py
megh-k: Sorry this was coppied from the whatsapp version of the test, I thought it was intentional

home/tests/test_api.py
```diff
@@ -92,6 +94,15 @@ def mk_test_doc() -> Document:
 
 @pytest.mark.django_db
 class TestContentPageAPI:
+    """
+    FIXME
+    ----------
+    Whatsapp body : Currently tests are split into whatsapp and not-whatsapp.
+        This is the moment the whatsapp body is different from the others such that
```
jerith, line 101: "This is the moment"?

jerith (dismissed):
I think we're good to go once that one docstring makes sense. 😄

jerith (approved):
🎉

PR #289, opened 2024-03-13
added platform env to Sentry config

Mudiwa66:


Mudiwa66:
I'm going to push this to the platform-mnch-contentrepo.prk-k8s.prd-p6t.org instance, so that environment is specific to platform.

PR #291, opened 2024-03-14
added sentry env to contentrepo config

Mudiwa66:
I added the sentry environment config to contentrepo, so now we can associate errors based on qa or platform instances of contentrepo

### DIFF
--- a/contentrepo/settings/production.py
+++ b/contentrepo/settings/production.py
@@ -18,6 +18,7 @@ ALLOWED_HOSTS = env.list("ALLOWED_HOSTS")
 CSRF_TRUSTED_ORIGINS = env.list("CSRF_TRUSTED_ORIGINS")
 
 SENTRY_DSN = env.str("SENTRY_DSN", "")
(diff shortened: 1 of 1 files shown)


contentrepo/settings/production.py
```diff
@@ -18,6 +18,7 @@
 CSRF_TRUSTED_ORIGINS = env.list("CSRF_TRUSTED_ORIGINS")
 
 SENTRY_DSN = env.str("SENTRY_DSN", "")
+SENTRY_ENVIRONMENT = env.str("SENTRY_ENVIRONMENT", "")
```
rudigiesler, line 21: The default environment is `production`, all lower case.

I think having it be an empty string is invalid, which makes it really easy to have an invalid configuration if you specify `SENTRY_DSN` but not `SENTRY_ENVIRONMENT`.

So I think it will be better to default this to `production` instead of an empty string.

Mudiwa66: I see, I've changed it. It should be okay now.

rudigiesler (approved):

PR #292, opened 2024-03-14
Use wagtail-modeladmin instead of deprecated contribs module

HawkiesZA:
## Purpose
ModelAdmin is deprecated in Wagtail. The recommendation is that we migrate to Snippets. However Snippets do not support Page models currently. In future there will be a migration path for "treeless listing views of pages".

## Approach
Instead of Migrating to Snippets for our ContentPageAdmin Model, we instead use the externally provided `wagtail-modeladmin` package.

## Learning

[Wagtail ModelAdmin to Snippet migration docs](https://docs.wagtail.org/en/v5.2.2/reference/contrib/modeladmin/migrating_to_snippets.html#using-modeladmin-to-manage-page-models)


rudigiesler (commented):
On line 137 of pyproject.toml, we're ignoring a warning about this issue. If we remove that warning filter, then we can check that we don't get that warning popping up.

Looks good otherwise :)

rudigiesler (dismissed):

rudigiesler (approved):

PR #295, opened 2024-03-18
Make a more informative message

HawkiesZA:
## Purpose
Importing messages with no page causes an error that isn't very informative, leaving the user thinking the system is broken.

## Approach
Change the message to be more informative so that the user has an idea of what to do to fix it.

### DIFF
--- a/home/import_content_pages.py
+++ b/home/import_content_pages.py
@@ -40,9 +40,17 @@ class ImportException(Exception):
     Base exception for all import related issues.
     """
 
-    def __init__(self, message: str, row_num: int | None = None):
+    def __init__(
+        self,
+        message: str,
+        row_num: int | None = None,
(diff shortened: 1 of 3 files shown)


rudigiesler (commented):
If I remember the ticket correctly, I think the issue here is with `add_message_to_shadow_content_page_from_row`, which calls `find_shadow_content_page`, which if it cannot find the page that the message is for, raises an error message `Cannot find content page with slug {slug} and locale {locale}`, which doesn't really help the user know what the issue is there.

Instead, we probably want an error like `This is a message for the content page with slug {slug} and locale {locale}, but no such page exists`, but only for the rows that are messages (we probably want a different but similar error for rows that are variation messages, I'm not sure if this function is used elsewhere)

rudigiesler (approved):

PR #296, opened 2024-03-19
Add content page delete validation

erikh360:
## Purpose
Block the delete of a page when it is linked in other places


home/apps.py
```diff
@@ -0,0 +1,12 @@
...
+
+class HomeAppConfig(AppConfig):
+    name = "home"
+
+    def ready(self):
+        from wagtail.models.reference_index import ReferenceIndex
+
+        from home.models import OrderedContentSet
+
+        ReferenceIndex.register_model(OrderedContentSet)
```
rudigiesler, line 12: Ah interesting, I thought that snippets were automatically registered: https://docs.wagtail.org/en/latest/advanced_topics/reference_index.html#configuration

home/models.py
```diff
@@ -839,6 +841,34 @@ def submit_whatsapp_template(self, previous_revision):
...
+                    url = reverse("wagtailadmin_pages:edit", args=(link.object_id,))
+                    page_links.append((url + tab, f"{page} - {link_type}"))
+
+                elif link.model_name == "Ordered Content Set":
+                    orderedcontentset = OrderedContentSet.objects.get(id=link.object_id)
+                    url = reverse(
+                        "wagtailsnippets_home_orderedcontentset:edit",
+                        args=(link.object_id,),
+                    )
+                    orderedcontentset_links.append((url, orderedcontentset.name))
```
rudigiesler, line 868: Should we have an `else` here that logs an error? So that we're at least aware if in the future someone adds an additional reference that's not covered by these two?

home/apps.py
erikh360: Yeah, looks like I can remove that.

erikh360:
@rudigiesler ready for another look please.

rudigiesler (approved):

PR #298, opened 2024-03-20
Omega 113 broken links script

Buhle79:
## Purpose

We have broken links for content pages and ordered content set due to pages that are deleted.

## Approach

Write a script to get all broken pages.

## Learning
_Describe the research stage_

#### Blog Posts


home/management/commands/broken_links_clean_up.py
```diff
@@ -0,0 +1,57 @@
...
+        ordered_content_links = []
+
+        # Only query live pages
+        pages = ContentPage.objects.live()
+
+        for page in pages:
+            if page.related_pages:
+                # Get related pages add it to the list if it is linked to a page that no longer exist
+                for rp in page.related_pages:
+                    if rp.value is not None:
```
rudigiesler, line 20: I'm not 100% sure how this works, but I think we should have a test for it anyway. My concern here is that `rp.value` still contains the ID to the related page, even when the page doesn't exist, and we have to actually lookup to see if a page with that ID still exists.

home/management/commands/broken_links_clean_up.py
```diff
@@ -0,0 +1,57 @@
...
+        for oc in ordered:
+            if oc.pages:
+                for page in oc.pages:
+                    if page.value.get("contentpage") is not None:
+                        continue
+                    ordered_content_links.append(
+                        {
+                            "Ordered Content": oc.id,
+                            "Error": "Page linked to it no longer exist",
+                        }
```
rudigiesler, line 52: Maybe we could instead just write these straight to stdout? That way each one appears on a separate line. I don't think we need this dictionary structure, it doesn't seem like we're using it, we just need to print out which ones of these have errors

home/management/commands/broken_links_clean_up.py
Buhle79: @rudigiesler it is possible that there is something that i'm doing wrong, page.related returns a length of 2 if one is deleted then rp.value for the one that is deleted is empty. But i will relook to confirm.

rudigiesler: Ah okay, maybe having a test to confirm will be best then? I could be wrong and `rp.value` does the database lookup for us under the hood

Buhle79:
Hi @rudigiesler for re look please.

home/tests/test_broken_links_clean_up.py
```diff
@@ -0,0 +1,130 @@
...
+        self_help_rp = Page.objects.get(pk=self_help.id)
+        PageBuilder.link_related(self.health_info, [self_help, index])
+
+        index.delete()
+        related_pages = [rp.value for rp in self.health_info.related_pages]
+        call_command("broken_links_clean_up", stdout=output)
+
+        assert len(self.health_info.related_pages) == 2
+        assert related_pages[1] is None
+        assert related_pages == [self_help_rp, None]
```
rudigiesler, line 83: I think we need an assert here that deals with the output of the command, to test that the command is giving us the correct list of broken links

home/tests/test_broken_links_clean_up.py
```diff
@@ -0,0 +1,130 @@
...
+
+        assert len(orig_self_help_btn) == 2
+        assert (
+            orig_self_help_btn[0]["value"]["title"]
+            == del_self_help_btn[0]["value"]["title"]
+        )
+        assert (
+            orig_self_help_btn[1]["value"]["page"]
+            == del_self_help_btn[1]["value"]["page"]
+        )
```
rudigiesler, line 130: Same here with asserting on the output

home/management/commands/broken_links_clean_up.py
```diff
@@ -0,0 +1,49 @@
...
+from home.models import ContentPage, OrderedContentSet
+
+
+class Command(BaseCommand):
+    help = "Clean broken links for pages"
+
+    def handle(self, *args, **options):
+
+        # Only query live pages
+        content_pages = ContentPage.objects.live()
```
rudigiesler, line 13: I think we want to query all pages, not just live pages. Otherwise we could have a page in draft with this issue, it goes unnoticed because this script doesn't pick it up, and then we push it live we have a broken live page.

rudigiesler (commented):
Just some comments around the tests. Also, we probably also want a test that deals with ordered content sets

Buhle79:
Hi @rudigiesler this PR is ready for re look please.

home/tests/test_broken_links_clean_up.py
```diff
@@ -0,0 +1,179 @@
...
+        assert output.getvalue().strip() == "Successfully retrieve broken links"
```
rudigiesler, line 138: This makes me think that this isn't working as expected? The output isn't listing the buttons pointing to deleted pages? But it might just be an issue with the test, it seems like we're calling the command before we delete the page, so if we call the command afterwards, then we should get the output we expect.

rudigiesler (commented):
Just a comment on the go to button test, I think it's a small fix though, otherwise looks good :)

home/tests/test_broken_links_clean_up.py
Buhle79: Oh yes thanks i see my mistake here.

rudigiesler (approved):

PR #299, opened 2024-03-20
Fix import orderedcontentset page

HawkiesZA:
## Purpose
When you click on the Import button for Ordered Content sets it throws a 500 error. This is because the dynamic URL changed from :index to :list.

## Approach
I changed the dynamic URL to use the :list url.


rudigiesler (dismissed):
I'm not sure if there's an easy way of unit testing this? Maybe just a GET and check we don't get a 500?

rudigiesler (approved):

PR #300, opened 2024-03-21
updated pr template

Mudiwa66:
## Purpose
We're redoing the pr template

### DIFF
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -1,16 +1,21 @@
+## Type
+- [ ] Feature
+- [ ] Bug
+- [ ] Update (fix something that did not work as expected)
+- [ ] Documentation
+
 ## Purpose
-_Describe the problem or feature in addition to a link to the issues._
+_Why does this exist?_
 
-## Approach
-_How does this change address the problem?_
+## Solution
+_What changed, and does this address the problem? _
 
-#### Open Questions and Pre-Merge TODOs
-- [ ] Use github checklists. When solved, check the box and explain the answer.
+## Tests
+_Was this tested? How?_
 
-## Learning
-_Describe the research stage_
+#### Checklist
+- [ ] Ran tests successfully
+- [ ] Errors handled or acknowledged
+- [ ] Added to release notes
+- [ ] Updated readme/documentation (if necessary)
 
-_Links to blog posts, patterns, libraries or addons used to solve this problem_
-
-#### Blog Posts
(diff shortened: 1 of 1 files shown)


pull_request_template.md
```diff
@@ -1,16 +1,21 @@
...
 
-#### Open Questions and Pre-Merge TODOs
-- [ ] Use github checklists. When solved, check the box and explain the answer.
+## Solution
+_What changed, and does this address the problem? _
 
-## Learning
-_Describe the research stage_
+## Tests
+_Was this tested? How?_
```
rudigiesler, line 14: I'm not too sure that we need a tests section, just having something on the checklist to make sure that we've remembered to add unit tests.

pull_request_template.md
```diff
@@ -1,16 +1,21 @@
...
 
-_Links to blog posts, patterns, libraries or addons used to solve this problem_
+#### Checklist
+- [ ] Ran tests successfully
+- [ ] Errors handled or acknowledged
```
rudigiesler, line 18: I think we can replace these two with just "Added or updated unit tests", since we'll want to make sure to do that for every change. The CI will make sure that all the tests are passing.

pull_request_template.md
```diff
@@ -1,16 +1,21 @@
-## Purpose
-_Describe the problem or feature in addition to a link to the issues._
+## Type
+- [ ] Feature
+- [ ] Bug
+- [ ] Update (fix something that did not work as expected)
+- [ ] Documentation
```
rudigiesler, line 5: Maybe we can make these rather match our changelog types? https://keepachangelog.com/en/1.1.0/

rudigiesler (commented):
Looks good! Just a few comments on how we could improve it a bit

pull_request_template.md
jerith: Agreed. The checklist should be for things that are sometimes forgotten and can't easily be checked in CI, and that includes adding/updating tests for all new/changed behaviour. CI will tell us that the tests passed, and error handling is more of a design/implementation thing that doesn't apply to all PRs.

pull_request_template.md
rudigiesler: If we're keeping the CHANGELOG up to date in the PR, then the type will be in there, so we can probably get rid of the type

Mudiwa66:
Is this better? Is there anything else I need to add? It looks a bit bare, but maybe something slimmer is better.

rudigiesler (approved):
Looks good! I'm fine with it being more bare, we can always add. The old one had too much, we were never using most of it, so better that it's leaner.

PR #303, opened 2024-03-25
Limit docs extensions

Hlamallama:
## Purpose
_Why does this exist?_
Validate document extensions that can be uploaded.

## Solution
This is a implementation that we didn't have in place.
It addressed the problem.

- [x] Added to release notes


contentrepo/settings/base.py
```diff
@@ -13,7 +13,8 @@
 DEFAULT_SECRET_KEY = "please-change-me"
 SECRET_KEY = os.environ.get("SECRET_KEY") or DEFAULT_SECRET_KEY
 ALLOWED_HOSTS = env.list("ALLOWED_HOSTS", default=["localhost"])
-
+WAGTAILDOCS_EXTENSIONS = ["doc", "docx", "xls", "xlsx", "ppt", "pptx", "pdf", "txt"]
+WAGTAILDOCS_DOCUMENT_MODEL = "home.CustomDocument"
```
rudigiesler, line 17: Looking at the documents, it doesn't seem like we need a whole custom document model to be able to use the `WAGTAILDOCS_EXTENSIONS` configuration, have you tried it without the custom model?

I'm worried about existing documents and how that's handled with a migration over to a new model.

Hlamallama: This was the only way I could get File extension “exe” is not allowed. Allowed extensions are: doc, docx, xls, xlsx, ppt, pptx, pdf, txt or else it will show like this File extension “” is not allowed. Allowed extensions are: doc, docx, xls, xlsx, ppt, pptx, pdf, txt.

rudigiesler: I did a test on my local environment, and applying this change removes all documents, so that's unfortunately not something we can apply and merge, otherwise we'll lose all documents in our production instances.

Before trying to add a migration to migrate things over to the new document model, however, this looks to me like a bug in Wagtail. 

So what I would suggest here is, to create a new blank project, using the latest version of wagtail (you can follow https://docs.wagtail.org/en/stable/getting_started/tutorial.html), and add the minimal amount to be able to upload documents, and see if you can replicate this issue. If you can, then we should create a detailed bug report on the wagtail repo, with all the reproduction information, as this is then something that needs to be fixed upstream in wagtail.

Hlamallama: Will have a look.
Thanks

contentrepo/settings/base.py
```diff
@@ -13,6 +13,7 @@
 DEFAULT_SECRET_KEY = "please-change-me"
 SECRET_KEY = os.environ.get("SECRET_KEY") or DEFAULT_SECRET_KEY
 ALLOWED_HOSTS = env.list("ALLOWED_HOSTS", default=["localhost"])
+WAGTAILDOCS_EXTENSIONS = ["doc", "docx", "xls", "xlsx", "ppt", "pptx", "pdf", "txt"]
```
rudigiesler, line 16: Maybe put a comment above the setting, saying that these are the extensions that whatsapp cloud API supports for documents (maybe with a link?), so that we know where we got this list from.

rudigiesler (commented):
I think we can bring that unit test back, that ensures we haven't made a mistake in the name of the config or anything, even if the assertion shows that incorrect message, at least we've documented that we know that the message is wrong and are happy with that until it gets fixed.

Then also just need to update the release notes :)

rudigiesler (approved):

PR #306, opened 2024-03-28
return status code 404 if page not found

Hlamallama:
## Purpose
Return status code for when a page doesn't exist instead of status code 400

## Solution
Change the exception from validation error to NotFound
Removed the i18n_patterns, this caused a redirect if there was no page found

#### Checklist
- [x] Added to release notes

### DIFF
--- a/contentrepo/urls.py
+++ b/contentrepo/urls.py
@@ -71,7 +71,7 @@ urlpatterns = urlpatterns + [
     #    path("pages/", include(wagtail_urls)),
 ]
 
-urlpatterns += i18n_patterns(
-    path("search/", search_views.search, name="search"),
-    path("", include(wagtail_urls)),
-)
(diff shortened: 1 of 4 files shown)


contentrepo/urls.py
```diff
@@ -71,7 +70,7 @@
...
 ]
 
-urlpatterns += i18n_patterns(
-    path("search/", search_views.search, name="search"),
-    path("", include(wagtail_urls)),
-)
+# urlpatterns += i18n_patterns(
+#     path("search/", search_views.search, name="search"),
+#     path("", include(wagtail_urls)),
+# )
```
jerith, line 76: Code that's commented out should either be deleted, or have an explanation of why it hasn't been deleted. In this case, I think let's just delete it.

jerith (approved):

PR #307, opened 2024-04-03
Manual template submission

fritzbrand:
## Purpose
We originally planned on making template submission manual, either by a button or a workflow, but 
after chatting with the content team, it was decided to keep the auto-submit functionality as is, i.e. no manual submission button or workflow process will be added.

## Solution
We will still add the new fields submission_name, submission_status, and submission_result to the interface, so the content team has more insight into the results of the submission process

#### Checklist
- [ ] Added to release notes (I'm not adding this to release notes yet, as its technically not ready for release yet. Will add when I add the details on standalone whatsapp templates


home/whatsapp.py
```diff
@@ -99,37 +99,131 @@ def from_locale(cls, locale: Locale) -> "WhatsAppLanguage":
...
+        headers=headers,
+        data=json.dumps(data, indent=4),
+    )
+
+    upload_details = {
+        "upload_session_id": response.json()["id"],
+        "upload_file": image_obj.file,
+    }
+
+    response.raise_for_status()
```
rudigiesler, line 133: We probably want to call `raise_for_status` before we call `.json()`, otherwise we're trying to decode the response before we start checking if it was a success.

home/whatsapp.py
```diff
@@ -99,37 +99,131 @@ def from_locale(cls, locale: Locale) -> "WhatsAppLanguage":
...
+    }
+    response = requests.post(
+        url,
+        headers=headers,
+        data=json.dumps(data, indent=4),
     )
-    if image_id is not None:
-        image_obj = Image.objects.get(id=image_id)
-        components.append(create_whatsapp_template_image(image_obj))
+    json_response_content = json.loads(response.content)
```
rudigiesler, line 199: Why not `response.json()` to be consistent everywhere else?

home/whatsapp.py
```diff
@@ -99,37 +99,131 @@ def from_locale(cls, locale: Locale) -> "WhatsAppLanguage":
...
+    result_string = ""
+    # check response code
+    if response.ok:
+        if json_response_content.get("id") is not None:
+            result_string = f"Template Submission OK. Template ID = {json_response_content.get('id')}"
+        return result_string
 
+    result_string = (
+        f"Template Submission ERROR. {response.json()['error']['error_user_msg']} "
+    )
```
rudigiesler, line 210: With this, we have no idea if the result of this function call was successful or not. Maybe this should raise an exception with a custom type that we can catch, with this as the exception message? That way we can determine if the call was successful or not, by whether or not it throws that exception.

home/whatsapp.py
```diff
@@ -183,97 +279,86 @@ def create_whatsapp_template_submission(
...
+
+        buttons = []
+        for button in quick_replies:
+            buttons.append({"type": "QUICK_REPLY", "text": button})
+        components.append({"type": "BUTTONS", "buttons": buttons})
+
+    return components
+
+
+def create_standalone_template_header_components(
```
rudigiesler, line 323: I think it's a little confusing that `create_standalone_template_body_components` returns a list of components, but this returns just a single component. Typing might help, but I feel it's too easy to make the mistake of calling `append` for both of these, and ending up with an invalid list of components.

Maybe we could just remove the `s` at the end of this function name, but that doesn't seem noticeable enough. 

So maybe the best option is for this function to also return a list, so that it's consistent, just that this function will only return a single item in that list.

home/models.py
```diff
@@ -1388,14 +1404,41 @@ def save_revision(
...
             # Log the error to sentry and send error message to the user
             logger.exception(f"Failed to submit template name:  {self.name}")
             raise ValidationError("Failed to submit template")
 
-        if template_name:
-            revision.content["name"] = template_name
+        if result_str:
+            revision.content["name"] = self.name
+            revision.content["submission_name"] = self.template_name
+            revision.content["submission_status"] = self.SubmissionStatus.SUBMITTED
```
rudigiesler, line 1440: Does this mean that whether there's an error or not, we're always setting the status to submitted? Should we not be using the failed status for errors?

home/templates/wagtailsnippets/snippets/home/orderedcontentset/index.html
```diff
@@ -4,7 +4,7 @@
 {% block import_button %}
     <div class="actionbutton">
             <div class="dropdown dropdown-button match-width col">
-                <a href="{% url 'import_orderedcontentset' %}" class="button bicolor button--icon">{% icon name="download" wrapped=1 %}{% trans 'Import' %}</a>
+                <a href="{% url 'import_orderedcontentset' %}" class="button bicolor button--icon">TEST1{% icon name="download" wrapped=1 %}{% trans 'Import' %}</a>
```
rudigiesler, line 7: I think we can remove this?

home/tests/test_models.py
```diff
@@ -730,7 +730,7 @@ def test_simple_template_submission(self) -> None:
         wat.save()
         wat.save_revision()
 
-        wat.submit_whatsapp_template(None)
+        # submit_whatsapp_template(wat)
```
rudigiesler, line 733: We can remove these comments, similar to how we removed the function call in the other tests.

home/tests/test_whatsapp.py
```diff
@@ -268,6 +279,30 @@ def test_create_standalone_whatsapp_template(self) -> None:
...
+            "language": "en_US",
+            "components": [{"type": "BODY", "text": "Test Body"}],
+        }
+        url = "http://whatsapp/graph/v14.0/27121231234/message_templates"
+        responses.add(responses.POST, url, json={})
+
+        locale = Locale.objects.get(language_code="en")
+        create_standalone_whatsapp_template(
+            "Test template 1", "Test Body", "UTILITY", locale=locale
+        )
```
rudigiesler, line 300: Should we maybe assert on the response from this function? So that we can test our success message?

home/tests/test_models.py
```diff
@@ -760,7 +760,6 @@ def test_template_create_with_buttons(self):
         wat.quick_replies.add(created_qr)
         wat.save()
         wat.save_revision()
-        wat.submit_whatsapp_template(None)
 
         request = responses.calls[0].request
         assert json.loads(request.body) == {
```
rudigiesler, line 765: Maybe we can also assert on that the correct status fields have been updates on the model?

home/tests/test_models.py
```diff
@@ -797,7 +796,6 @@ def test_template_create_with_example_values(self):
         )
         wat.save()
         wat.save_revision()
-        wat.submit_whatsapp_template(None)
 
         request = responses.calls[0].request
```
rudigiesler, line 800: Do we have a test here for what happens when the call to whatsapp fails? If not, then we should probably add one

home/whatsapp.py
fritzbrand: 👍

home/whatsapp.py
fritzbrand: 👍

home/models.py
fritzbrand: Sorry, I meant to come back to that, but somehow forgot

rudigiesler (approved):

PR #260, opened 2024-02-21
Add data model for assessments

rudigiesler:
## Purpose
We want to be able to store the structure of assessments in ContentRepo

## Approach
We're going to approach this similarly to ordered content sets, where we add an additional Assessment model to the database to manage the assessments

#### Open Questions and Pre-Merge TODOs
- [x] Add data model
- [x] Add assessments to wagtail admin
- [x] Add assessments API
- [x] Add assessments import/export
- [ ] Update documentation

PR #309, opened 2024-04-04
Assessment Import-Export

DevChima:
## Purpose
This PR adds the assessment page from this [pr](https://github.com/praekeltfoundation/contentrepo/pull/260) to ContentRepo along with an xls/csv exporter function for exporting assessments.

- [ ] Exporter 
- [ ] Importer


rudigiesler:
@DevChima It's rather difficult to review this, as it contains all the changes of the other branch as well. I'm going to try change the base of this branch to that branch, to see if it can help the diff

rudigiesler:
I've tried changing the base to `add-assessments`, but this seems to have changes from main that that branch didn't. So I merged main into `add-assessments`, but there's still merge conflict issues between that branch and this one. I'll take a look next week

home/export_assessments.py
```diff
@@ -0,0 +1,258 @@
...
+            answers = []
+            scores = []
+            for answer in question.value["answers"]:
+                answers.append(answer["answer"])
+                scores.append(str(answer["score"]))
+            question_data.append(
+                {
+                    "questions": (question.value["question"]).replace("\n", ","),
+                    "error": (question.value["error"]),
+                    "answers": ", ".join(answers),
```
rudigiesler, line 94: What if the answers have commas in them? We can probably use the `csv` library here, which will properly escape the commas, instead of a `", ".join`

home/export_assessments.py
```diff
@@ -0,0 +1,258 @@
...
+        question_data = []
+        for question in questions:
+            answers = []
+            scores = []
+            for answer in question.value["answers"]:
+                answers.append(answer["answer"])
+                scores.append(str(answer["score"]))
+            question_data.append(
+                {
+                    "questions": (question.value["question"]).replace("\n", ","),
```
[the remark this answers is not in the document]
rudigiesler: Also, this is a single question, maybe the title is better as `question` instead of `questions`?

home/export_assessments.py
```diff
@@ -0,0 +1,258 @@
...
+    menu_style.fill = blue_fill
+    menu_style.font = Font(bold=True, size=10)
+
+    # Add named styles to wb
+    wb.add_named_style(header_style)
+    wb.add_named_style(menu_style)
+
+    # column widths
+
+    # Set menu style for any "Menu" row
```
rudigiesler, line 234: I don't think we have any `Menu` rows in the assessment exports, so we can get rid of this styling

home/export_assessments.py
```diff
@@ -0,0 +1,258 @@
...
+    rows: list[dict[str | int, str | int]]
+
+    def __init__(self, queryset: PageQuerySet):
+        self.rows = []
+        self.queryset = queryset
+
+    def perform_export(self) -> list[dict[str | int, str | int]]:
+        for item in self.queryset:
+            questions_data = self.get_questions(item.questions)
+            for i, question_data in enumerate(questions_data):
```
rudigiesler, line 60: This means that if an assessment doesn't have any questions, then it won't be included in the export.

Instead, maybe we can export the assessment as one row, and then the questions for that assessment as separate rows underneath that?

home/export_assessments.py
```diff
@@ -0,0 +1,258 @@
...
+            scores = []
+            for answer in question.value["answers"]:
+                answers.append(answer["answer"])
+                scores.append(str(answer["score"]))
+            question_data.append(
+                {
+                    "questions": (question.value["question"]).replace("\n", ","),
+                    "error": (question.value["error"]),
+                    "answers": ", ".join(answers),
+                    "score": ", ".join(scores),
```
rudigiesler, line 95: Should this maybe be `scores`? Since it's all the scores for all the answers

home/templates/wagtailsnippets/snippets/index.html
```diff
@@ -21,7 +21,7 @@
             {% block import_button %}
                 <div class="actionbutton">
                         <div class="dropdown dropdown-button match-width col">
-                            <a href="{% url 'import' %}" class="button bicolor button--icon">{% icon name="download" wrapped=1 %}{% trans 'Import' %}</a>
+                            <a href="{% url 'import_assessment' %}" class="button bicolor button--icon">{% icon name="download" wrapped=1 %}{% trans 'Import' %}</a>
```
rudigiesler, line 24: Won't this change it for all snippets? ie. it will also change it for ordered content sets, not just for assessments

home/export_assessments.py
```diff
@@ -0,0 +1,258 @@
...
+
+    def perform_export(self) -> list[dict[str | int, str | int]]:
+        for item in self.queryset:
+            questions_data = self.get_questions(item.questions)
+            for i, question_data in enumerate(questions_data):
+
+                self.rows.append(
+                    {
+                        "title": item.title,
+                        "page_id": item.id,
```
rudigiesler, line 65: This isn't a page ID, this is an assessment. And I don't think we need the assessment ID in the export, we use the slug and locale as the unique identifier here.

home/tests/test_assessment_import_export.py
```diff
@@ -0,0 +1,473 @@
...
+@filter_both
+def ignore_certain_fields(entry: ExpDict) -> ExpDict:
+    # FIXME: Do we need page.id to be imported? At the moment nothing in the
+    #        import reads that.
+    # FIXME: Implement import/export for doc_link, image_link, media_link.
+    ignored_fields = {
+        "page_id",
+        "doc_link",
+        "image_link",
+        "media_link",
```
rudigiesler, line 57: We don't have any documents, images, or media for assessments, so we don't need these.

home/export_assessments.py
DevChima: @rudigiesler Would this be necessary since we can't have an assessment without a question. I understand questions are mandatory.

home/export_assessments.py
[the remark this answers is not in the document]
DevChima: It was a test that somehow snuck through. I'll remove.

home/export_assessments.py
rudigiesler: Ah, yes they are, okay we don't need to separate it out then, since you need at least 1 question

home/import_assessment_pages.py
```diff
@@ -0,0 +1,361 @@
...
+    locale: str = ""
+    high_result_page: list[str] = field(default_factory=list)
+    high_inflection: str = ""
+    medium_result_page: list[str] = field(default_factory=list)
+    medium_inflection: str = ""
+    low_result_page: list[str] = field(default_factory=list)
+    low_inflection: str = ""
+    generic_error: str = ""
+    question_count: str = ""
+    question: list[str] | list[dict[str, Any]] | str = field(default_factory=list)
```
rudigiesler, line 250: Why would this be a list of strings? Surely every row only contains a single question?

home/import_assessment_pages.py
```diff
@@ -0,0 +1,361 @@
...
+                medium_result_page=grouped_rows_list[0].medium_result_page,
+                medium_inflection=grouped_rows_list[0].medium_inflection,
+                low_result_page=grouped_rows_list[0].low_result_page,
+                low_inflection=grouped_rows_list[0].low_inflection,
+                generic_error=grouped_rows_list[0].generic_error,
+                question_count=str(len(questions)),
+                question=questions,
+                answers=[],
+                scores=[],
+                error="",
```
rudigiesler, line 317: I think this should be `grouped_rows_list[0].error`

home/import_assessment_pages.py
```diff
@@ -0,0 +1,361 @@
...
+                high_inflection=grouped_rows_list[0].high_inflection,
+                medium_result_page=grouped_rows_list[0].medium_result_page,
+                medium_inflection=grouped_rows_list[0].medium_inflection,
+                low_result_page=grouped_rows_list[0].low_result_page,
+                low_inflection=grouped_rows_list[0].low_inflection,
+                generic_error=grouped_rows_list[0].generic_error,
+                question_count=str(len(questions)),
+                question=questions,
+                answers=[],
+                scores=[],
```
rudigiesler, line 316: Why are we putting the question, answers, and scores all into the `question` field, and then leaving `answers` and `scores` blank? 

I think the problem here is that we're trying to use `ContentRow`, which is a single row in the spreadsheet, to represent a `ShadowAssessmentPage`, which is made up of a collection of rows in a spreadsheet, one for each question. 

I think this function should rather be taking in `ContentRow`s, grouping those rows, and returning `ShadowAssessmentPage`s

home/import_assessment_pages.py
```diff
@@ -0,0 +1,361 @@
...
+
+    return output_rows
+
+
+def result_page_get_id_from_slug(slug: list[str]) -> int:
+    try:
+        page = ContentPage.objects.get(slug=slug)
+    except ObjectDoesNotExist:
+        raise ImportAssessmentException(
+            f"You are trying to add an assessment with slug {slug} that does not exist. Please create the {slug} page first."
```
rudigiesler, line 329: It's not the assessment's slug that does not exist, it's one of the result page's slugs that doesn't exist.

So maybe a better wording for this is `You are trying to add an assessment, where one of the result pages references the content page with slug {slug}, which does not exist. Please create the content page first`

home/export_assessments.py
```diff
@@ -0,0 +1,257 @@
...
+                        "slug": item.slug,
+                        "locale": str(item.locale.language_code),
+                        "high_result_page": str(item.high_result_page.slug),
+                        "high_inflection": item.high_inflection,
+                        "medium_result_page": str(item.medium_result_page.slug),
+                        "medium_inflection": item.medium_inflection,
+                        "low_result_page": str(item.low_result_page.slug),
+                        "low_inflection": 0,
+                        "generic_error": item.generic_error,
+                        "question_count": i + 1,
```
rudigiesler, line 75: I'm unsure what this `question_count` is? It looks to be like it's the index of the answer for this specific question? In that case, maybe `answer_index` is a better name for it?

It also looks like we're not looking at it at all in the import. So to avoid confusion, I would suggest either removing it, or making the import order the answers for each question by this field.

home/tests/test_assessment_import_export.py
```diff
@@ -0,0 +1,512 @@
...
+    def ff(src: ExpDict, dst: ExpDict) -> ExpPair:
+        return filter_func(src), filter_func(dst)
+
+    return ff
+
+
+@filter_both
+def strip_leading_whitespace(entry: ExpDict) -> ExpDict:
+    # FIXME: Do we expect imported content to have leading spaces removed?
+    bodies = {k: v.lstrip(" ") for k, v in entry.items() if k.endswith("_body")}
```
rudigiesler, line 51: We don't have any fields that end with `_body` for assessments, so it seems like this won't do anything, so we can probably remove it from the assessment tests?

home/tests/test_assessment_import_export.py
```diff
@@ -0,0 +1,512 @@
...
+            bodies=[
+                WABody("Low Score", [WABlk("*Low Inflection Page")]),
+                MBody("Low Score", [MBlk("Low Inflection Page")]),
+            ],
+        )
+
+        csv_bytes = csv_impexp.import_file("comma_separated_answers.csv")
+        content = csv_impexp.export_assessment()
+        src, dst = csv_impexp.csvs2dicts(csv_bytes, content)
+        assert dst == src
```
rudigiesler, line 436: I think we can add another test here to deal with newlines in the question, I don't see that being tested anywhere.

rudigiesler:
@DevChima Happy for you to merge :)

PR #313, opened 2024-04-16
Use data from excel; don't append row if no data

HawkiesZA:
## Purpose
We are creating a new spreadsheet that will act as the Single Source of Truth for content between Figma and ContentRepo. This content is used in Figma and needs to be imported to ContentRepo.

## Solution
1. When importing Excel files in ContentRepo, formulae are imported as strings instead of the values they represent. This makes it so that we at least use the last cached data from the last time the spreadsheet was opened.
2. We try to create ContentRows even if there's no row data. This fixes the issue.

#### Checklist
- [ ] Added to release notes


home/import_content_pages.py
```diff
@@ -224,7 +226,8 @@ def clean_excel_cell(cell_value: str | float | datetime | None) -> str:
             for name, cell in zip(header, row):  # noqa: B905 (TODO: strict?)
                 if name and cell:
                     r[name] = clean_excel_cell(cell)
-            rows.append(ContentRow.from_flat(r))
+            if r:
+                rows.append(ContentRow.from_flat(r))
```
rudigiesler, line 230: Does this work? Maybe we need a test? I'm expecting `r` here to still be `{"title": "", ...}` for blank rows, unless blank rows get returned as `[]`?

rudigiesler (commented):
Looks good! Just a question around skipping blank rows

home/import_content_pages.py
HawkiesZA: Added a test :-)

rudigiesler (dismissed):

PR #315, opened 2024-04-23
added ordered content set tests

Mudiwa66:
## Purpose
To increase testing for the CMS

## Solution
Added tests for ordered content sets.
#### Checklist


home/tests/test_models.py
```diff
@@ -495,6 +496,35 @@ def test_status_live_plus_draft(self):
         ordered_content_set.save_revision()
         self.assertEqual(ordered_content_set.status(), "Live + Draft")
 
+    def test_get_relationship(self):
+        ordered_content_set = OrderedContentSet(name="Test Title")
```
jerith, line 500: Can you add explanatory docstrings to all these tests, please? (Look at some of the existing tests (particularly around import/export and views) for examples.)

Mudiwa66: Good idea, I've added them.

Mudiwa66: My bad, fixed.

home/tests/test_models.py
```diff
@@ -465,36 +466,90 @@ def test_template_create_with_pt_language(self, mock_create_whatsapp_template):
...
+        ordered_content_set.profile_fields.append(("age", "15-18"))
+        ordered_content_set.save()
+        self.assertEqual(ordered_content_set.get_age(), "15-18")
+
+    def test_get_page(self):
+        """
+        Ordered Content Sets with an page selected should return the choosen page. We compare
+        the unique slug of a page
+        """
```
jerith, line 536: Will they always return the first page?

Mudiwa66: It returns a list of pages, I was just checking specifically for the first page since I only imported one.

jerith: Perhaps a better test here would be to have multiple pages and assert that `.page()` (which is badly named, apparently) returns all their slugs in order?

We probably also want a test for the case where we don't have any pages.

Mudiwa66: Good idea, I've added that.

jerith (dismissed):

rudigiesler (approved):

PR #317, opened 2024-04-25
Template import export

fritzbrand:
## Purpose
We need to be able to export and import the new Standalone WhatsApp Templates

## Solution
Built custom import and export functionality, based on ContentPages and Assessments

#### Checklist
- [ ] Added to release notes

Not adding this to docs or readme, will form part of larger Standalone Templates docs and readme when that is ready


home/export_whatsapp_templates.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def to_tuple(self) -> tuple[str | int, ...]:
+        return astuple(self)
+
+
+class WhatsAppTemplateExporter:
+    rows: list[dict[str | int, str | int]]
+
+    def __init__(self, queryset: PageQuerySet):
+        self.rows = []
```
rudigiesler, line 44: I don't think we're using `rows`, so we can remove it

home/export_whatsapp_templates.py
```diff
@@ -0,0 +1,188 @@
...
+        self.queryset = queryset
+
+    def perform_export(self) -> Iterable[ExportRow]:
+        image_link = ""
```
rudigiesler, line 48: This, being outside the loop, will mean that it will be blank, until we come across a template with an image, at which point it will have a value. But then any further templates will keep the value, until another image comes along to replace that value. So if you have templates like:
```
Template(image=None)
Template(image="url1")
Template(image=None)
Template(image="url2")
```
You'll get:
```
""
"url1"
"url1"
"url2"
```
Instead of:
```
""
"url1"
""
"url2"
```

I think we just need to move this into the for loop to fix it.

home/export_whatsapp_templates.py
```diff
@@ -0,0 +1,188 @@
...
+                example_values=serialize_list(
+                    [v["value"] for v in item.example_values.raw_data]
+                ),
+                submission_name=str(item.submission_name),
+                submission_status=str(item.submission_status),
+                submission_result=(item.submission_result),
+            )
+
+
+def filter_non_empty(items: Iterable[str]) -> Iterator[str]:
```
rudigiesler, line 68: It doesn't seem like we're using this function, so we can get rid of it

home/import_whatsapp_templates.py
```diff
@@ -0,0 +1,205 @@
...
+        WhatsAppTemplate.objects.all().delete()
+
+    def default_locale(self) -> Locale:
+        site = Site.objects.get(is_default_site=True)
+        return site.root_page.locale
+
+
+@dataclass(slots=True, frozen=True)
+class ContentRow:
+    template_id: int | None = None
```
rudigiesler, line 147: We're not exporting the template ID, and we're not using it in the import either, so we can probably remove this field.

home/import_whatsapp_templates.py
```diff
@@ -0,0 +1,205 @@
...
+            quick_replies=deserialise_list(row.pop("quick_replies", "")),
+            example_values=deserialise_list(row.pop("example_values", "")),
+            submission_name=str(row.pop("submission_name", "")),
+            submission_status=str(row.pop("submission_status", "")),
+            submission_result=str(row.pop("submission_result", "")),
+            **row,
+        )
+
+
+def deserialise_dict(value: str) -> dict[str, str]:
```
rudigiesler, line 180: It doesn't seem like we're using this, so we can remove it

home/models.py
```diff
@@ -1466,6 +1454,26 @@ class SubmissionStatus(models.TextChoices):
...
+    submission_status = models.CharField(
+        max_length=30,
+        choices=SubmissionStatus.choices,
+        blank=True,
+        default=SubmissionStatus.NOT_SUBMITTED_YET,
+    )
+
+    submission_result = models.TextField(
+        help_text="The result of submitting the template",
+        null=True,
```
rudigiesler, line 1466: I know this was here before, but we probably don't want this to be nullable as well as blank, since it's a string type: https://docs.djangoproject.com/en/5.0/ref/models/fields/#null

home/templates/wagtailsnippets/snippets/index.html
```diff
@@ -22,7 +22,9 @@
                 <div class="actionbutton">
                         <div class="dropdown dropdown-button match-width col">
                             <a href="{% url 'import' %}" class="button bicolor button--icon">{% icon name="download" wrapped=1 %}{% trans 'Import' %}</a>
+                            <a href="{% url 'import_assessment' %}" class="button bicolor button--icon">{% icon name="download" wrapped=1 %}{% trans 'Import' %}</a>
```
rudigiesler, line 25: This seems unrelated, maybe a merge issue? We can probably remove it

home/tests/test_whatsapp_template_import_export.py
```diff
@@ -0,0 +1,372 @@
...
+    def ff(src: ExpDict, dst: ExpDict) -> ExpPair:
+        return filter_func(src), filter_func(dst)
+
+    return ff
+
+
+@filter_both
+def strip_leading_whitespace(entry: ExpDict) -> ExpDict:
+    # FIXME: Do we expect imported content to have leading spaces removed?
+    bodies = {k: v.lstrip(" ") for k, v in entry.items() if k.endswith("_body")}
```
rudigiesler, line 49: We don't have any fields in templates that end with `_body`, so I don't think this filter will do anything, so maybe we don't need it?

home/tests/test_whatsapp_template_import_export.py
```diff
@@ -0,0 +1,372 @@
...
+        fsrcs.append(src)
+        fdsts.append(dst)
+    return fsrcs, fdsts
+
+
+def csv2dicts(csv_bytes: bytes) -> ExpDicts:
+    return list(csv.DictReader(StringIO(csv_bytes.decode())))
+
+
+WEB_PARA_RE = re.compile(r'^<div class="block-paragraph">(.*)</div>$')
```
rudigiesler, line 72: It doesn't look like this is being used

rudigiesler (commented):
Just a few comments, mostly things that we're not using that we can remove. I also see things that have been commented out, we can probably remove those few things as well

rudigiesler (approved):

PR #318, opened 2024-04-30
added max char variation msg import validation

Mudiwa66:
## Purpose
To increase import validation and error handling

#### Checklist
- [x ] Added or updated unit tests


Mudiwa66:
@jerith I'm able to get both variation error messages but I can't return a list of errors for all rows

jerith:
> @jerith I'm able to get both variation error messages but I can't return a list of errors for all rows

Just one row is fine. Collecting errors for multiple rows is a lot trickier, because we assume all rows prior to the one currently being imported have succeeded.

home/import_content_pages.py
```diff
@@ -505,15 +506,30 @@ def errors_to_strings(self, errs: dict[str, list[str]]) -> str | list[str]:
...
+                        if key == "messages":
+                            messages.append(value)
+                        else:
+                            messages.extend(extract_messages(value))
+                return messages
+
+            for val in json_data_errors["blockErrors"][0]["blockErrors"].values():
+                print(f"errors: {val}")
+                messages = extract_messages(val)
+                error_messages.append(messages)
```
jerith, line 529: Should this perhaps be `.extend()`?

jerith:
Okay, I think I've figured this out. The test CSV now has a missing parent, which raises an `ImportException` before we reach the form-based validation. In that case we'll only get the missing parent error, which is fine. What we care about in this PR is returning multiple validation errors, which is possible because we get them all together in one big data structure.

Mudiwa66:
I think this works as expected now.

home/import_content_pages.py
```diff
@@ -53,13 +54,13 @@ class ImportException(Exception):
 
     def __init__(
         self,
-        message: str,
+        message: str | list[Any],
```
jerith, line 57: I think this can be `str | list[str]`.

home/import_content_pages.py
```diff
@@ -491,34 +494,60 @@ def validate_page_using_form(self, page: Page) -> None:
...
             # TODO: better error stuff
             if errs:
-                error_message = self.errors_to_strings(errs)
-                raise ImportException(
-                    f"Validation error: {error_message}", self.row_num
-                )
+                errors = []
+                error_message = self.errors_to_list(errs)
+                for err in error_message:
+                    errors.extend(ValidationError(f"Validation error: {err}"))
```
jerith, line 500: Do we want this to be a `ValidationError`? We're expecting strings in `ImportException`. Also, since we're not passing in an iterable, we need `.append()` rather than `.extend()`.

home/tests/test_content_import_export.py
```diff
@@ -710,17 +710,17 @@ def test_no_translation_key_nondefault(self, csv_impexp: ImportExport) -> None:
 
         assert e.value.row_num == 4
         # FIXME: Find a better way to represent this.
-        assert "translation_key" in e.value.message
-        assert "“” is not a valid UUID." in e.value.message
+        assert "translation_key" in e.value.message[0]
+        assert "“” is not a valid UUID." in e.value.message[0]
```
jerith, line 714: Instead of just looking at the first message, let's rather assert on the whole list. If that's impractical (perhaps if there's a substring we don't know ahead of time) we can extract it with `[msg] = e.value.message` to make sure there's only one.

home/import_content_pages.py
```diff
@@ -281,8 +281,11 @@ def create_content_page_index_from_row(self, row: "ContentRow") -> None:
                 self.home_page(locale).add_child(instance=index)
             index.save_revision().publish()
         except ValidationError as err:
-            # FIXME: Find a better way to represent this.
-            raise ImportException(f"Validation error: {err}")
+            for key in err.error_dict:
+                field_name = key
+            raise ImportException(
+                f"Validation error: {field_name} - {'; '.join(err.messages)}"
```
jerith, line 287: If we're getting more than one of these, perhaps add them each as a separate message like we do with the wagtail validation errors?

jerith (approved):

PR #323, opened 2024-05-09
Page view report test

Hlamallama:
## Purpose
_Why does this exist?_

## Solution

#### Checklist
- [ ] Added to release notes


home/tests/test_views.py
```diff
@@ -688,3 +691,115 @@ def test_ordered_form_has_all_expected_options(self, admin_client):
...
+            page_id=content_page.id,
+            revision_id=content_page.get_latest_revision().id,
+            platform="whatsapp",
+        )
+
+        filterset_data = {"page": content_page.id}
+        filter_set = PageViewFilterSet(filterset_data, queryset=PageView.objects.all())
+        filtered_queryset = filter_set.qs
+
+        assert filtered_queryset.count() == 3
```
rudigiesler, line 781: I'm not sure that this test is actually testing anything. We created a single page, only put pageviews on that page, and then filtering on that single page? I think for this test, we probably want to make two pages, have pageviews on both, and then ensure that when we filter by page we only get the pageviews for the one page and not the other.

rudigiesler (commented):
Just one of the tests don't quite look right to me, the rest look great :)

rudigiesler (approved):

PR #326, opened 2024-05-30
sample solution

Mudiwa66:
## Purpose
To return a useful error when the JSON button is incorrect.

## Solution
Made a sample solution, still in progress.

#### Checklist
- [X] Added or updated unit tests


home/import_content_pages.py
```diff
@@ -232,18 +233,20 @@ def clean_excel_cell(cell_value: str | float | datetime | None) -> str:
         first_row = next(worksheet.iter_rows(max_row=1, values_only=True))
         header = [clean_excel_cell(cell) if cell else None for cell in first_row]
         rows: list[ContentRow] = []
+        i = 1
```
jerith, line 236: We need to start with row 2 here. Row 0 doesn't exist (we start numbering at 1) and row 1 is the headers. (And similarly in `parse_csv()` below.)

jerith (approved):

PR #332, opened 2024-06-06
Add multiselect question type

HawkiesZA:
## Purpose
We use different styles of questions which also require different types of validation. Forms needs to cater for all the question types that we require.

## Solution
This adds the ability to add Multiselect Questions to a Form

#### Checklist
- [x] Added to release notes


DevChima (approved):
👍

PR #335, opened 2024-06-11
Add Semantic ID

HawkiesZA:
## Purpose
We need a semantic ID for each answer so that we can save the ids of the answers.

## Solution
Added a semantic ID to categorical and multiple choice answers

#### Checklist
- [x] Added to release notes


DevChima (approved):
👍

PR #338, opened 2024-06-20
Add integer type question with min and max fields

DevChima:
## Purpose
We need the ability to add a integer question type to a Form with min / max validation

## Solution
This adds integer type question to the form and a min/max value which we can use for validation.

#### Checklist
- [x] Added to release notes


HawkiesZA (approved):
:+1:

PR #340, opened 2024-06-27
optional inflection, results pages and tests

DevChima:
## Purpose
Results pages are not required for every question type so it can be optional.

## Solution
Make inflection points and results pages optional in the model and cover for attribute errors

#### Checklist
- [ ] Added to release notes


HawkiesZA (approved):
:+1:

PR #342, opened 2024-07-02
Add validation for IntegerQuestionBlock min and max

DevChima:
## Purpose
We need some validation to enforce rules for IntegerQuestionBlock forms so that at form level, min and max values are not the same and min is not greater than max. 

## Solution
This pr or adds custom model validation to handle this.

#### Checklist
- [x] Added to release notes

### DIFF
--- a/home/tests/test_models.py
+++ b/home/tests/test_models.py
@@ -25,6 +25,7 @@ from home.models import (
     ContentPageIndex,
     GoToPageButton,
     HomePage,
+    IntegerQuestionBlock,
     NextMessageButton,
     OrderedContentSet,
     PageView,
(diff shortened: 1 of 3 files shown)


home/tests/test_models.py
```diff
@@ -953,3 +954,32 @@ def test_template_create_failed(self):
...
+class IntegerQuestionBlockTests(TestCase):
+    def create_min_max_value(
+        self,
+        min,
+        max,
+    ):
+        return {
+            "min": min,
+            "max": max,
+        }
```
erikh360, line 968: This looks like more code than just sending a dict directly to clean(), any reason for this?

DevChima: No reason in particular. I just felt it was cleaner to do this with a function instead of sending the dict directly as
IntegerQuestionBlock().clean({
"min": 40,
"max": 50
}

erikh360 (approved):

PR #343, opened 2024-07-03
custom model validation for min and max to only accept positive values

DevChima:
## Purpose
We do not need IntegerQuestion type to hold negative min and max values as there's no real world use case for it in this form.

## Solution
Add custom validation to make sure that negative vaues for min and max are not saved.

#### Checklist
- [ ] Added to release notes

### DIFF
--- a/home/tests/test_models.py
+++ b/home/tests/test_models.py
@@ -992,7 +992,7 @@ class IntegerQuestionBlockTests(TestCase):
         with self.assertRaises(ValidationError) as e:
             IntegerQuestionBlock().clean(self.create_min_max_value(min=50, max=-40))
         self.assertEqual(
-            "min cannot be greater than max",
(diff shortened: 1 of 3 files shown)


home/tests/test_models.py
```diff
@@ -983,3 +983,9 @@ def test_clean_identical_min_max(self):
             "min cannot be greater than max",
             e.exception.message,
         )
+        with self.assertRaises(ValidationError) as e:
+            IntegerQuestionBlock().clean(self.create_min_max_value(min=-50, max=40))
```
erikh360, line 987: I would add another separate assert where max is negative too.

Right now the code on the other side of the OR is untested.

erikh360 (approved):
Looks good! Happy for you to merge after adding the extra assert

erikh360 (approved):

PR #352, opened 2024-08-01
dynamically accessed error level

Mudiwa66:
## Purpose
The import errors are not showing up for all types of errors

## Solution
The  error level is now dynamically accessed during error extraction.


jerith (changes requested):
Can you add a test for this, please? Specifically, one that fails with the old implementation but passes with the new one.

home/tests/import-export-data/max chars button.csv
```diff
@@ -0,0 +1,1362 @@
+no,web_title,slug generator,slug,Domain,parent,whatsapp_title,Image,Media,whatsapp_body,footer,tags,ButtonA,Button A type,ButtonB,ButtonC,ListOption,OptionA,OptionB,OptionC,OptionD,OptionE,OptionF,OptionG,OptionH,OptionI,OptionJ,CMS buttons,A label cms,A destination cms,B button label cms,B destination cms,C button label cms,C destination cms,D button label cms,D destination cms,E button label cms,E destination cms,F button label cms,F destination cms,G button label cms,G destination cms,H button label cms,H destination cms,I button label cms,I destination cms,J button label cms,J destination cms,A json code,B json code,C json code,D json code,E json code,F json code,G json code,H json code,I json code,J json code,buttons,locale,whatsapp_template_category
```
jerith, line 1: Can we try trim this file down to something more manageable? 😅

home/tests/test_content_import_export.py
```diff
@@ -1083,6 +1083,19 @@ def test_invalid_JSON_button(self, csv_impexp: ImportExport) -> None:
         assert e.value.row_num == 3
         assert e.value.message == ["Bad JSON button, you have: Broken Button"]
 
+    @pytest.mark.xfail(reason="This will fail because of the old error implementation")
```
jerith, line 1086: Now that we've seen it fail, let's put the fix back and make sure it passes. 🙂

PR #354, opened 2024-08-12
Fix submission status

fritzbrand:
## Purpose
There are some bugs on the submissions status fields on standalone templates, noteably

- Mismatch between status fields in Edit vs List view
- Incorrect fields used in List view
- Unhandled exception on publish of template
- Update max chat limit after Facebook changes

## Solution

#### Checklist
- [ ] Added to release notes


rudigiesler (commented):
I've got some concerns around this. We shouldn't be publishing when saving a revision, since that means you can't have anything in draft. We also shouldn't be overwriting revision history, we'll want to be able to see in the history the templates that failed submission.

Also, are there not any unit tests around this? Shouldn't we be adding checks that this submission is updating all the fields that we expect it to?

PR #356, opened 2024-08-19
Ordered content export

Mudiwa66:
## Purpose
Ordered content exports do no work as intended

## Solution
Ordered content sets now have their own explicit export code, and can now export in XLSX format.

#### Checklist
- [X ] Added or updated unit tests


home/export_ordered_sets.py
```diff
@@ -0,0 +1,172 @@
...
+                profile_field=item.profile_field,
+                page_slugs=item.page_slugs,
+                page=item.page,
+                unit=item.unit,
+                before_or_after=str(item.before_or_after),
+                contact_field=item.contact_field,
+            )
+
+
+def serialize_list(items: Iterable[str]) -> str:
```
rudigiesler, line 64: We don't seem to be using this function anywhere

home/export_ordered_sets.py
```diff
@@ -0,0 +1,172 @@
...
+        "generic_error": 370,
+        "question": 370,
+        "explainer": 370,
+        "error": 370,
+        "min": 110,
+        "max": 110,
+        "answers": 370,
+        "scores": 110,
+        "answer_semantic_ids": 110,
+        "question_semantic_id": 110,
```
rudigiesler, line 129: Most of these columns don't appear in ordered content sets. The ones that aren't in ordered content sets should be removed, and the missing ones added.

home/export_ordered_sets.py
```diff
@@ -0,0 +1,172 @@
...
+    for column in sheet.iter_cols(max_row=1):
+        [cell] = column
+        width = column_widths_in_pts[str(cell.value)]
+        sheet.column_dimensions[get_column_letter(cell.col_idx)].width = ceil(
+            width / adjustment
+        )
+
+    # Freeze heading row and side panel, 1 added because it freezes before the column
+    panel_column = get_column_letter(5)
+    sheet.freeze_panes = sheet[f"{panel_column}2"]
```
rudigiesler, line 140: I don't think there's any side panels in this, I think this will just freeze the name and profile fields. I don't think we need any frozen panels for ordered content sets, it's not a very wide table.

home/export_ordered_sets.py
```diff
@@ -0,0 +1,172 @@
...
+        return asdict(self)
+
+    def to_tuple(self) -> tuple[str]:
+        return astuple(self)
+
+
+class OrderedSetExporter:
+    rows: list[ExportRow]
+
+    def __init__(self, queryset: PageQuerySet):
```
rudigiesler, line 47: I don't think this is a PageQuerySet, since this is a query set of ordered content sets.

home/export_ordered_sets.py
```diff
@@ -0,0 +1,172 @@
...
+    wb.add_named_style(header_style)
+
+    # Set header style for row 1
+    for row in sheet["1:2"]:
+        for cell in row:
+            cell.style = header_style
+
+    # Set dividing border for side panel
+    for cell in sheet[f"{panel_column}:{panel_column}"]:
+        cell.border = left_border
```
rudigiesler, line 161: Same here, this will just add a dividing border after profile fields.

home/tests/test_content_import_export.py
```diff
@@ -1303,6 +1316,29 @@ class TestExport:
...
+        """
+        Ordered Content Sets should export in XLSX format correctly
+        """
+        ordered_content_set = OrderedContentSet(name="Test Title")
+        ordered_content_set.save()
+
+        # exports in xlsx format
+        content = xlsx_impexp.export_ordered_content()
+
+        assert content is not None
```
rudigiesler, line 1340: I think we need more comprehensive testing than this. We're not testing that the export is actually outputting sensible data that reflects what is in the database. And we're also not testing that if we export ordered content sets, that they'll be the same when we import them again.

home/export_ordered_sets.py
Mudiwa66: Should I create a custom query set for these ordered sets?

rudigiesler: I don't think we have proper typing for django, but could you just use the base `QuerySet` as a return type?

Mudiwa66:
Closing this, moving to [370](https://github.com/praekeltfoundation/contentrepo/pull/370)

PR #369, opened 2024-08-28
Fix detail url

HawkiesZA:
## Purpose
The detail_url in the Pages API is currently showing the current URL and not the detail URL, which causes a failure on one of our implementations.

## Solution
We use `get_object_detail_url` to get the correct detail url for the Page. Trying to use super() and building on that doesn't work, as it seems to cause a cascade of errors when Assessments serialize a Page.

#### Checklist
- [x] Added to release notes

### DIFF
--- a/home/tests/test_api.py
+++ b/home/tests/test_api.py
@@ -505,6 +505,42 @@ class TestContentPageAPI:
             "has_children": False,
         }
 
+    @pytest.mark.parametrize("platform", ALL_PLATFORMS)
+    def test_detail_view_meta(self, uclient, platform):
+        """
+        Fetching the detail view of a page returns the page metadata.
+        """
+        page = self.create_content_page(tags=["self_help"], body_type=platform)
+        response = uclient.get(f"/api/v2/pages/{page.id}/")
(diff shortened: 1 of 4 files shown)


home/tests/test_api.py
```diff
@@ -2120,6 +2219,10 @@ def test_assessment_detail_endpoint_filter_by_tag(self, uclient):
         assert meta["slug"] == self.low_result_page.slug
         assert meta["parent"]["id"] == self.low_result_page.get_parent().id
         assert meta["locale"] == "en"
+        assert (
+            meta["detail_url"]
+            == f"http://localhost/api/v2/pages/{self.low_result_page.id}/"
+        )
```
rudigiesler, line 2225: Could we maybe have one test that just asserts the entire response? That way we can make sure none of the other meta fields have incorrect values, and we can ensure that future changes don't change the API?

HawkiesZA: Done

rudigiesler (approved):

rudigiesler (approved):

PR #373, opened 2024-09-04
Updated Standalone Whatsapp Templates  list column titles and values

fritzbrand:
## Purpose
QA logged the following two issues:

- After you save a template with correct info the page submission status shows as Submitted. but on the home page it shows as NOT_SUBMITTED_YET
- Also can we format this text to not have underscores on the templates home page and have the text look like all other statuses in content repo.

## Solution
- I check with Rudi and this is the standard Wagtail behaviour.  The list view always displays the most recent published revision, while the edit screen shows the latest revision available.  You should see that ContentPages behave the same
- Have updated the list column titles and values to be more in line with the rest of the list views

#### Checklist
- [NA] Added or updated unit tests
- [NA] Added to release notes
- [NA] Updated readme/documentation (if necessary)

### DIFF
--- a/home/models.py
+++ b/home/models.py
@@ -1656,8 +1656,7 @@ class WhatsAppTemplate(
             revision.content["submission_name"] = self.template_name
             revision.content["submission_status"] = self.SubmissionStatus.SUBMITTED
             revision.content["submission_result"] = (
-                # f"Success! Template ID = {response_json['id']}"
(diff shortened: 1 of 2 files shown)


HawkiesZA (approved):
LGTM. I'm guessing there isn't an easy way to add tests?

fritzbrand:
Yea, this was mostly a change in the UI. Thx

PR #374, opened 2024-09-05
Bumped version to 1.2.2

fritzbrand:
## Purpose
To deploy v1.2.2

## Solution
Fixed column titles and values for standalone whatsapp templates

#### Checklist
- [x] Added to release notes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Updated idna from 3.6 to 3.7
 -->
 
+## v1.2.2
+### Added
+- Standalone Whatsapp Templates: Fixed list display column titles and values
+
+
 ## v1.2.1
 ### Added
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -25,6 +25,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Updated idna from 3.6 to 3.7
 -->
 
+## v1.2.2
+### Added
+- Standalone Whatsapp Templates: Fixed list display column titles and values
```
rudigiesler, line 30: What about all the other changes in unreleased? They'll also be included in any release we create. That will probably be a major version bump.

Also, we probably want things to be QAd before releasing, unless we have a good reason to release untested changes?

fritzbrand: Sorry I got my wires twisted here.  I was trying to deploy this to QA so Meimoen can test next week, but I should probably have deployed `1.2.2-dev.0` to QA, rather than making a new version

PR #375, opened 2024-09-18
Add forms model validation to import

HawkiesZA:
## Purpose
When we import Forms without required fields, the UI doesn't display an error. However when you then go into the imported Form and try to save it, you get an error that required fields are not filled.

## Solution
Similar to content pages (in fact we reuse the code originally created for content pages), we now apply model validation on import.

#### Checklist
- [x] Added to release notes


home/models.py
```diff
@@ -1266,7 +1266,7 @@ class BaseQuestionBlock(blocks.StructBlock):
         required=False,
         help_text="Error message for this question if we don't understand the input",
     )
-    semantic_id = blocks.TextBlock(help_text="Semantic ID for this question")
+    question_semantic_id = blocks.TextBlock(help_text="Semantic ID for this question")
```
rudigiesler, line 1269: Why were these changes needed?

home/tests/test_content_import_export.py
```diff
@@ -1329,7 +1329,7 @@ def test_invalid_page_already_in_db(self, csv_impexp: ImportExport) -> None:
         assert e.value.row_num == 3
         # FIXME: Find a better way to represent this.
         assert e.value.message == [
-            "Validation error: example_values - The number of example values provided (1) does not match the number of variables used in the template (3)"
+            "Validation error at row 3: example_values - The number of example values provided (1) does not match the number of variables used in the template (3)"
```
rudigiesler, line 1332: Do we need the row number added here? The row number is already included in the Exception, and the code that converts the Exception to an error message for the user adds the row number to the message, so now we'll have duplicate row numbers in the error message.

rudigiesler (commented):
Just a question around changing the semantic_id fields, since this changes the API, which is probably something that we don't want to do unnecessarity. If we could avoid doing that it would be great. Also renaming the field creates a migration, and I'm not sure if that migration will wipe existing data or if it will just rename the field, if we want to rename the field we'll want tests to ensure that the migration isn't erasing data (the wagtail streamfield migrations are a bit tricky: https://docs.wagtail.org/en/v4.2/advanced_topics/streamfield_migrations.html)

home/import_helpers.py
```diff
@@ -0,0 +1,159 @@
...
+        locale: Locale | None = None,
+    ):
+        self.row_num = row_num
+        self.message = [message] if isinstance(message, str) else message
+        self.slug = slug
+        self.locale = locale
+        super().__init__()
+
+
+def wagtail_to_formdata(val: Any) -> Any:
```
jerith, line 41: 🎉 🎉 🎉 🎉

rudigiesler (approved):

PR #379, opened 2024-10-22
Fix Ordered Content import

HawkiesZA:
## Purpose
When we import Ordered Sets we aren't taking into account that the columns "Time", "Unit", "Before or After", or "Contact Page" could contain a comma separated string of values.

## Solution
This PR takes the above into account, and also adds some validation for if the number of items in those columns don't match up.

#### Checklist
- [x] Added to release notes

### DIFF
--- a/home/import_ordered_content_sets.py
+++ b/home/import_ordered_content_sets.py
@@ -88,11 +88,19 @@ class OrderedContentSetImporter:
         )
 
         if (
-            len(times) != 0
-            and len(times) != len(units)
-            or len(times) != len(before_or_afters)
-            or len(times) != len(page_slugs)
-            or len(times) != len(contact_fields)
+            len(
+                {
+                    len(item)
+                    for item in [
+                        times,
+                        units,
+                        before_or_afters,
+                        page_slugs,
+                        contact_fields,
+                    ]
+                }
+            )
+            != 1
         ):
             raise ImportException(
                 f"Row {row['Name']} has {len(times)} times, {len(units)} units, {len(before_or_afters)} before_or_afters, {len(page_slugs)} page_slugs and {len(contact_fields)} contact_fields and they should all be equal.",
@@ -136,24 +144,25 @@ class OrderedContentSetImporter:
         """
         ordered_set.pages = []
         for page in pages:
-            if not page.page_slug or page.page_slug == "-":
-                continue
-            content_page = ContentPage.objects.filter(slug=page.page_slug).first()
-            if page:
-                ordered_set.pages.append(
-                    (
-                        "pages",
-                        {
-                            "contentpage": content_page,
-                            "time": page.time or "",
-                            "unit": page.unit or "",
-                            "before_or_after": page.before_or_after or "",
-                            "contact_field": page.contact_field or "",
-                        },
+            if page.page_slug and page.page_slug != "-":
+                content_page = ContentPage.objects.filter(slug=page.page_slug).first()
+                if content_page:
+                    ordered_set.pages.append(
+                        (
+                            "pages",
+                            {
+                                "contentpage": content_page,
+                                "time": page.time or "",
+                                "unit": page.unit or "",
+                                "before_or_after": page.before_or_after or "",
+                                "contact_field": page.contact_field or "",
+                            },
+                        )
                     )
-                )
-            else:
-                logger.warning(f"Content page not found for slug '{page.page_slug}'")
+                else:
+                    raise ImportException(
+                        f"Content page not found for slug '{page.page_slug}'"
+                    )
 
     def _create_ordered_set_from_row(
         self, index: int, row: dict[str, str]
--- a/home/import_ordered_content_sets.py
+++ b/home/import_ordered_content_sets.py
@@ -32,7 +32,7 @@ class OrderedContentSetImporter:
         self.progress_queue = progress_queue
 
     def _get_or_init_ordered_content_set(
-        self, row: dict, set_name: str
+        self, row: dict[str, str], set_name: str
     ) -> OrderedContentSet:
         """
         Get or initialize an instance of OrderedContentSet from a row of a CSV file.
@@ -67,12 +67,12 @@ class OrderedContentSetImporter:
     def _validate_extracted_values(
         self,
         index: int,
-        row: dict,
-        times: list,
-        units: list,
-        before_or_afters: list,
-        page_slugs: list,
-        contact_fields: list,
+        row: dict[str, str],
+        times: list[str],
+        units: list[str],
+        before_or_afters: list[str],
+        page_slugs: list[str],
+        contact_fields: list[str],
     ) -> None:
         """
         Validate that the extracted values (times, units, before_or_afters, page_slugs, contact_fields) from a row of the file have the same length.
@@ -94,11 +94,11 @@ class OrderedContentSetImporter:
     def _create_pages(
         self,
         ordered_set: OrderedContentSet,
-        times: list,
-        units: list,
-        before_or_afters: list,
-        page_slugs: list,
-        contact_fields: list,
+        times: list[str],
+        units: list[str],
+        before_or_afters: list[str],
+        page_slugs: list[str],
+        contact_fields: list[str],
     ) -> None:
         """
         Given the extracted values from a row of the file, create the corresponding ordered content set pages.
@@ -136,7 +136,9 @@ class OrderedContentSetImporter:
             else:
                 logger.warning(f"Content page not found for slug '{page_slug}'")
 
-    def _create_ordered_set_from_row(self, index: int, row: dict) -> OrderedContentSet:
+    def _create_ordered_set_from_row(
+        self, index: int, row: dict[str, str]
+    ) -> OrderedContentSet:
         """
         Create an ordered content set from a single row of the file.
 
@@ -170,7 +172,7 @@ class OrderedContentSetImporter:
         ordered_set.save()
         return ordered_set
 
-    def perform_import(self):
+    def perform_import(self) -> None:
         """
         Import ordered content sets from a file.
 
--- a/home/import_ordered_content_sets.py
+++ b/home/import_ordered_content_sets.py
@@ -144,18 +144,14 @@ class OrderedContentSetImporter:
             if page.page_slug and page.page_slug != "-":
                 content_page = ContentPage.objects.filter(slug=page.page_slug).first()
                 if content_page:
-                    ordered_set.pages.append(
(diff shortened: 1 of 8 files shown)


jerith (commented):
My screen's not big enough to show enough enough of the code at once, which makes it _really_ difficult to review. Can we restructure this gigantic function-with-another-function-inside-it into a class similar to the content page import?

jerith:
> Can we restructure this gigantic function-with-another-function-inside-it into a class similar to the content page import?

If not, I'll try make time to work through it and understand what's going on later today. Otherwise someone else will have to review. Sorry. 😕

home/import_ordered_content_sets.py
```diff
@@ -0,0 +1,227 @@
...
+        # backwards compatiblilty if there's only one contact field
+        contact_fields = row["Contact Field"].split(",")
+        contact_fields = (
+            [p.strip() for p in contact_fields]
+            if len(contact_fields) > 1
+            else [contact_fields[0]] * len(times)
+        )
+
+        self._validate_extracted_values(
+            index, row, times, units, before_or_afters, page_slugs, contact_fields
```
jerith, line 165: Instead of passing these around separately a bunch of times, could we make a class to hold them and move all the logic related to parsing and such onto that?

home/import_ordered_content_sets.py
```diff
@@ -0,0 +1,227 @@
...
+        ordered content sets by extracting relevant information from each row.
+        The progress of the import operation is reported via a queue.
+
+        Raises:
+            ImportException: If any inconsistencies are found in the data
+            row while creating ordered content sets.
+        """
+        file = self.file.read()
+        lines = []
+        if self.filetype == "XLSX":
```
jerith, line 193: Can we separate the CSV-specific and XLSX-specific bits into their own methods that both return a list of rows? That way we don't have to worry about the details of parsing each file type in this method.

home/import_ordered_content_sets.py
```diff
@@ -0,0 +1,227 @@
...
+            for dictionary in reader:
+                lines.append(dictionary)
+
+        # 10% progress for loading file
+        self._set_progress(10)
+
+        for index, row in enumerate(lines):  # type: ignore
+            os = self._create_ordered_set_from_row(index, row)  # type: ignore
+            if not os:
+                print(f"Ordered Content Set not created for row {index + 1}")
```
jerith, line 225: Should we log this and/or make it visible in the UI rather than printing it?

home/tests/test_content_import_export.py
```diff
@@ -1185,6 +1185,21 @@ def test_multiple_field_errors(self, csv_impexp: ImportExport) -> None:
...
+            csv_impexp.import_ordered_sets(content)
+
+        assert e.value.row_num == 0
+        assert e.value.message == [
+            "Row Test Set has 2 times, 2 units, 3 before_or_afters, 3 page_slugs and 3 contact_fields and they should all be equal."
+        ]
+
     def test_import_ordered_sets_csv(self, csv_impexp: ImportExport) -> None:
         """
         Importing a CSV file with ordered content sets should not break
```
jerith, line 1205: This seems like a pretty low bar, and the test appears to do more than that. Can we update the docstring to reflect that?

jerith (commented):
I've reviewed the importer as if it's new code (because it's new to me) and left a bunch of comments/questions. (I only looked at new/updated tests, though.)

home/import_ordered_content_sets.py
```diff
@@ -0,0 +1,227 @@
...
+        self, row: dict[str, str], set_name: str
+    ) -> OrderedContentSet:
+        """
+        Get or initialize an instance of OrderedContentSet from a row of a CSV file.
+
+        :param row: The row of the CSV file, as a dict.
+        :param set_name: The name of the ordered content set.
+        :return: An instance of OrderedContentSet.
+        """
+        ordered_set = OrderedContentSet.objects.filter(name=set_name).first()
```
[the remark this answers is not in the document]
HawkiesZA: That's a good question. It sounds like a Rudi question though (and possibly a broader content question). Names are probably expected to be unique, but not enforced.

home/import_ordered_content_sets.py
```diff
@@ -0,0 +1,227 @@
...
+        if not ordered_set:
+            ordered_set = OrderedContentSet(name=set_name)
+
+        ordered_set.profile_fields = []
```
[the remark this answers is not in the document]
HawkiesZA: I think an empty list turns into an empty StreamField.

home/import_ordered_content_sets.py
```diff
@@ -0,0 +1,227 @@
...
+        return ordered_set
+
+    def _csv_to_list(self, column: str) -> list[str]:
+        """
+        Return a list of strings parsed from a column of a CSV row.
+
+        :param column: The column to parse.
+        :return: A list of strings, split from the value of row[column].
+        """
+        return [p.strip() for p in column.split(",")]
```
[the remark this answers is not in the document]
HawkiesZA: In this case I think everything is a string

home/import_ordered_content_sets.py
```diff
@@ -0,0 +1,227 @@
...
+            or len(times) != len(before_or_afters)
+            or len(times) != len(page_slugs)
+            or len(times) != len(contact_fields)
+        ):
+            raise ImportException(
+                f"Row {row['Name']} has {len(times)} times, {len(units)} units, {len(before_or_afters)} before_or_afters, {len(page_slugs)} page_slugs and {len(contact_fields)} contact_fields and they should all be equal.",
+                index,
+            )
+
+    def _create_pages(
```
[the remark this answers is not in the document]
HawkiesZA: Adding.

home/import_ordered_content_sets.py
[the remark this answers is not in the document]
jerith: Can we make a note of this somewhere so we can come back and revisit it?

home/import_ordered_content_sets.py
```diff
@@ -0,0 +1,253 @@
...
+        """
+        Add profile fields to an instance of OrderedContentSet from a row of a CSV file.
+
+        :param ordered_set: An instance of OrderedContentSet.
+        :param row: The row of the CSV file, as a dict.
+        """
+        ordered_set.profile_fields = []
+        for field in [f.strip() for f in (row["Profile Fields"] or "").split(",")]:
+            if field and field != "-":
+                [field_name, field_value] = field.split(":")
```
jerith, line 72: Are we guaranteed to have exactly one `:`? If not, can we maybe raise an `ImportException` with a suitable message if we have the wrong number?

home/import_ordered_content_sets.py
```diff
@@ -0,0 +1,253 @@
...
+        units = self._csv_to_list(row["Unit"])
+        before_or_afters = self._csv_to_list(row["Before Or After"])
+        page_slugs = self._csv_to_list(row["Page Slugs"])
+        # backwards compatiblilty if there's only one contact field
+        contact_fields = row["Contact Field"].split(",")
+        contact_fields = (
+            [p.strip() for p in contact_fields]
+            if len(contact_fields) > 1
+            else [contact_fields[0]] * len(times)
+        )
```
jerith, line 88: I find this a little hard to parse and understand. Also, it looks like we don't `.strip()` the fields if we only have one -- is that intentional?
Perhaps we can make the single vs multiple thing a little more distinct:
```suggestion
        contact_fields = self._csv_to_list(row["Contact Field"])
        # backwards compatiblilty if there's only one contact field
        if len(contact_fields) == 1:
            contact_fields = [contact_fields[0]] * len(times)
```

home/import_ordered_content_sets.py
HawkiesZA: This would be stuff like gender:male, so yes

HawkiesZA: Added a ticket for enforcing uniqueness in names: https://praekelt.leankit.com/card/31512228558429

jerith: > This would be stuff like gender:male, so yes

That would be a requirement and expectation, not a guarantee. Perhaps we could use some tests for `gender::male` and `gender-male` and such to make sure we're reporting content bugs instead of failing with no useful feedback to the user?

home/import_ordered_content_sets.py
```diff
@@ -0,0 +1,227 @@
...
+        :param times: The times extracted from the row.
+        :param units: The units extracted from the row.
+        :param before_or_afters: The before or afters extracted from the row.
+        :param page_slugs: The page slugs extracted from the row.
+        :param contact_fields: The contact fields extracted from the row.
+        """
+        for idx, page_slug in enumerate(page_slugs):
+            if not page_slug or page_slug == "-":
+                continue
+            page = ContentPage.objects.filter(slug=page_slug).first()
```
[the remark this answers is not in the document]
HawkiesZA: > We're not filtering the pages by locale here, so we may get multiple with the same slug. Are ordered content sets supposed to be per-locale?

Ordered Sets don't have a locale from what I can tell and the import format doesn't specify a locale for the page either. I'm going to log this as a bug that needs investigating.

home/import_ordered_content_sets.py
```diff
@@ -133,24 +141,25 @@ def _add_pages(
...
+                            "pages",
+                            {
+                                "contentpage": content_page,
+                                "time": page.time or "",
+                                "unit": page.unit or "",
+                                "before_or_after": page.before_or_after or "",
+                                "contact_field": page.contact_field or "",
+                            },
+                        )
+                    )
```
jerith, line 158: This is very verbose. Can we build the dict first and avoid most of the brackets-only lines?
```suggestion
                    os_page = {
                        "contentpage": content_page,
                        "time": page.time or "",
                        "unit": page.unit or "",
                        "before_or_after": page.before_or_after or "",
                        "contact_field": page.contact_field or "",
                    }
                    ordered_set.pages.append(("pages", os_page))
```

home/import_ordered_content_sets.py
```diff
@@ -0,0 +1,259 @@
...
+                        times,
+                        units,
+                        before_or_afters,
+                        page_slugs,
+                        contact_fields,
+                    ]
+                }
+            )
+            != 1
+        ):
```
jerith, line 101: A local variable could _really_ improve readability here:
```suggestion
        fields = [times, units, before_or_afters, page_slugs, contact_fields]
        if len({len(item) for item in fields}) != 1:
```

jerith (approved):

PR #382, opened 2024-10-29
Add list_button_title field

HawkiesZA:
## Purpose
Currently we have to hardcode list titles in Journeys because CMS doesn't have a way to store them.

## Solution
This adds a new field to the model to support the list button title.

#### Checklist
- [x] Added to release notes


home/models.py
```diff
@@ -272,6 +272,11 @@ class WhatsappBlock(blocks.StructBlock):
         required=False,
         max_num=3,
     )
+    list_button_title = blocks.CharBlock(
```
jerith, line 275: To avoid confusion with buttons, which are different from lists, perhaps `list_title` would be a better name?

(If we decide to make this change, we should do the same in the import/export code, etc.)

home/tests/test_api.py
```diff
@@ -947,7 +951,9 @@ def test_list_items(self, uclient):
         """
         test that list items are present in the whatsapp message
         """
-        page = self.create_content_page(list_items=["list item 1", "list item 2"])
+        page = self.create_content_page(
+            list_button_title="List Button", list_items=["list item 1", "list item 2"]
+        )
```
jerith, line 956: Do we have a test for a list with no title?

HawkiesZA: Added one

home/models.py
HawkiesZA: Changed

jerith (approved):

PR #383, opened 2024-11-05
Improve list item actions

HawkiesZA:
## Purpose
For WhatsApp messages, we have two kinds of interactive messages, button and list types.

For button messages, we currently have two kinds of buttons, next message buttons and go to page buttons. The next message button goes to the next whatsapp message in the current content page, and the go to page button goes to the first message in the linked content page (which is also configured in the CMS)

For list message items, we are currently only able to specify the text. We would like to also be able to specify the action along with the item text, similar to button messages.

## Solution
This allows us to specify the action for list message items.

#### Checklist
- [x] Added to release notes


home/migrations/0083_alter_contentpage_whatsapp_body.py
```diff
@@ -0,0 +1,262 @@
+# Generated by Django 4.2.11 on 2024-11-04 14:59
```
rudigiesler, line 1: Have you tested this migration? I have a feeling that it will just delete all list items, since it doesn't know to default them to `next_message` list items. I think we need a manual migration step as part of this: https://docs.wagtail.org/en/v4.2/advanced_topics/streamfield_migrations.html#streamfield-data-migrations

home/models.py
```diff
@@ -277,15 +292,12 @@ class WhatsappBlock(blocks.StructBlock):
         help_text="List title, up to 24 characters.",
         max_length=24,
     )
-    list_items = blocks.ListBlock(
-        blocks.CharBlock(label="Title"),
-        default=[],
+    list_items = blocks.StreamBlock(
+        [("next_message", NextMessageListItem()), ("go_to_page", GoToPageListItem())],
         help_text="List item title, up to 24 characters.",
```
rudigiesler, line 297: We should probably change this help text as well

home/models.py
```diff
@@ -323,7 +335,7 @@ def clean(self, value):
                 f"{len(result['message'])} characters long"
             )
 
-        list_items = result["list_items"]
+        list_items = [item.value["title"] for item in result["list_items"]]
```
rudigiesler, line 338: We should test if we still need this validation, and maybe remove it if not? With the new error translation put in place for import, I think this is no longer necessary, since we specify the max char length on the streamfield definition.

home/tests/import-export-data/list_items_with_comma_output.csv
```diff
@@ -0,0 +1,6 @@
...
+Sub 1.1.1,1,6,list-item-test,Test 1,List item test,,,List item test,message test,,UTILITY,,,,List Title,"[{""type"": ""next_message"", ""title"": ""Item 3""}, {""type"": ""next_message"", ""title"": ""Item 4""}, {""type"": ""next_message"", ""title"": ""Item 2""}, {""type"": ""next_message"", ""title"": ""Item 1""}]",,,,,,,,,f351c9f0-751c-466e-acb6-09d607410491,,,,English,,"[{""type"": ""next_message"", ""title"": ""Btn 1""}, {""type"": ""next_message"", ""title"": ""Btn 2""}]",,,,,
+Sub 1.1.1.1,1,10,importexport,List item test,import/export Test,,,import/export test,import/export test,,UTILITY,,,,List Title,"[{""type"": ""next_message"", ""title"": ""item 2, with comma""}, {""type"": ""next_message"", ""title"": ""item 3""}, {""type"": ""next_message"", ""title"": ""item 1""}, {""type"": ""next_message"", ""title"": ""Item' comma""}]",,,,,,,,,cd9f38ec-6cd9-4289-873a-417d996493b4,,,,English,,[],,,,,
+Sub 1.1.1.2,1,11,test-import,List item test,Test import,,,Test import,Test import,,UTILITY,,,,List Title,"[{""type"": ""next_message"", ""title"": ""item1""}, {""type"": ""next_message"", ""title"": ""item2""}]",,,,,,,,,6387db5b-ed76-4d34-86f4-92fbd9e1ebda,,,,English,,[],,,,,
```
rudigiesler, line 6: I think we should have at least one test that uses the `go_to_page` list item type.

home/migrations/0083_alter_contentpage_whatsapp_body.py
HawkiesZA: So the data isn't dropped, but it doesn't actually show up (e.g. if you undo the migration, everything goes back to normal).

I did try to do a manual migration step, but it never seemed to actually take. My thinking was that as a fallback we could do an export before upgrading the CMS version, then do an import after the upgrade.

Should I try the manual data migration step again? Maybe this time you can help solve it :-)

home/models.py
HawkiesZA: What should it be? "List item text, up to 24 characters"?

rudigiesler: Well this is now just the items in the list right? It's no longer just the text, each text block will have their own help text mentioning the text and character limit. Maybe "Items to appear in the list message"

home/tests/import-export-data/list_items_with_comma_output.csv
HawkiesZA: ha, and adding that test revealed that I hadn't fully implemented the logic for that!

home/models.py
HawkiesZA: ohhh, right, with you now

rudigiesler (approved):

PR #385, opened 2024-11-19
Fix list items

HawkiesZA:
## Purpose
The previous migration didn't actually run the stream data migration because it didn't have a null, so we need to do it again and also convert nulls to empty lists in the process.

## Solution
This should fix the broken QA build

#### Checklist
- [ ] Added to release notes


home/migrations/0083_alter_contentpage_whatsapp_body.py
```diff
@@ -31,7 +31,7 @@ def list_string_to_action(block_value):
 
 class ListStringToActionOperation(BaseBlockOperation):
     def apply(self, block_value):
-        list_string_to_action(block_value)
+        list_string_to_action(block_value)  # this is wrong, there should be a return
```
rudigiesler, line 34: Does this not imply data loss then? Should we maybe fix it before it gets deployed to production?

home/migrations/0084_alter_contentpage_whatsapp_body.py
```diff
@@ -0,0 +1,51 @@
...
+from wagtail.blocks.migrations.utils import formatted_list_child_generator
+from django.db import migrations
+
+
+def list_string_to_action(block_value):
+    # formatted_list_child_generator requires at least one item
+    if not block_value:
+        return []
+    new_list = []
+    for list_item in formatted_list_child_generator(block_value):
```
rudigiesler, line 15: These items will now be in streamfield format, not list format, so I think we'll have issues running it through twice. I think this only works now because all the values were set to `null` from the previous migration. Maybe we just have:
```
if not block_value:
    return []
return block value
```
To set any null values to an empty list?

rudigiesler (approved):

PR #387, opened 2024-11-21
Submit Buttons when creating a template

KaitCrawford:
## Purpose
Without this we are unable to use Buttons with template messages because they not submitted. However, we want to maintain the old functionality of submitting (sorted) quick replies if no buttons are provided. 

## Solution
On template submission we now submit Buttons if there are any present or Quick Replies if there are no Buttons. This means that existing pages should not have both Buttons and Quick Replies if the templates are expected to use the Quick Replies

#### Checklist
- [ /] Added or updated unit tests
- [ ] Added to release notes

### DIFF
--- a/home/models.py
+++ b/home/models.py
@@ -807,13 +807,22 @@ class ContentPage(UniqueSlugMixin, Page, ContentImportMixin):
         return self.quick_reply_items.all().values_list("tag__name", flat=True)
 
     @property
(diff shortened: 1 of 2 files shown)


rudigiesler (approved):
Looks good! I'm going to suggest holding off on merging for now, just because we're almost done with QA for a bunch of work, and want to make a release first, just to get those features out early next week.

PR #393, opened 2024-11-28
Add slugs locale orderedset

HawkiesZA:
## Purpose
The migration fails on QA because we're not setting a default locale for ordered sets.

## Solution
This sets a default locale for ordered sets and also fixes the slug deduplication (which did not take into account that all slugs will be empty at the start).

#### Checklist
- [ ] Added to release notes


home/migrations/0085_orderedcontentset_locale_orderedcontentset_slug.py
```diff
@@ -3,47 +3,53 @@
...
-            candidate_slug = slug
-            while OrderedContentSet.objects.filter(
-                slug=candidate_slug, locale=locale
-            ).exists():
+    for slug in duplicate_slugs:
+        ordered_content_sets = OrderedContentSet.objects.filter(slug=slug)
+        for ordered_content_set in ordered_content_sets:
+            slug_from_name = (
+                ordered_content_set.name.lower().replace(" ", "-").replace("_", "-")
+            )
```
rudigiesler, line 22: Maybe we can use Django's `slugify` function here? https://docs.djangoproject.com/en/5.1/ref/utils/#django.utils.text.slugify . I think slugs are allowed to have underscores

home/tests/test_migrations.py
```diff
@@ -53,32 +53,22 @@ def test_deduplicates_orderedcontentset_slugs(self) -> None:
...
-        )
-        OrderedContentSet.objects.create(
-            name="unique", slug="unique", locale=default_locale
-        )
+        # Create ordered content sets with blank slugs
+        OrderedContentSet.objects.create(name="first", slug="", locale=default_locale)
+        OrderedContentSet.objects.create(name="first", slug="", locale=default_locale)
+        OrderedContentSet.objects.create(name="first_pt", slug="", locale=pt_locale)
+        OrderedContentSet.objects.create(name="third", slug="", locale=default_locale)
+        OrderedContentSet.objects.create(name="unique", slug="", locale=default_locale)
```
rudigiesler, line 61: Maybe we can have some of the names here with whitespace and repeated dashes, to ensure that we're converting name -> slug correctly?

rudigiesler (commented):
Just a small thing with how we convert name to slug that I think we can improve, otherwise looks good!

home/migrations/0085_orderedcontentset_locale_orderedcontentset_slug.py
HawkiesZA: Didn't know that existed. Thanks!

home/tests/test_migrations.py
jerith: And mixed case?

HawkiesZA: slugify lowercases

rudigiesler (approved):

PR #397, opened 2024-12-04
Add fix rows

HawkiesZA:
## Purpose
1. We have import code for different models that do the same thing. 
2. Headers in import files are case sensitive

## Solution
1. We consolidate the shared code into our helpers
2. We added a helper function that lowercases headers internally so that it no longer matters what gets supplied.

#### Checklist
- [x] Added to release notes


home/import_helpers.py
```diff
@@ -157,3 +163,55 @@ def extract_errors(data: dict[str | int, Any] | list[str]) -> dict[str, str]:
...
+def fix_rows(rows: Generator[dict[str, str], None, None]) -> Iterator[dict[str, str]]:
+    """
+    Fix keys for all rows by lowercasing and removing whitespace
+    """
+    for row in rows:
+        yield fix_row(row)
+
+
+def fix_row(row: dict[str, str]) -> dict[str, str]:
+    return {k.lower().strip(): v.strip() if v else v for k, v in row.items()}
```
rudigiesler, line 177: The `if v else v` is confusing me here, I'm not sure it does anything? What are we trying to do here?

home/import_helpers.py
```diff
@@ -157,3 +163,55 @@ def extract_errors(data: dict[str | int, Any] | list[str]) -> dict[str, str]:
...
+    file_content: bytes, file_type: str
+) -> Iterator[tuple[int, dict[str, Any]]]:
+    read_rows = read_xlsx if file_type == "XLSX" else read_csv
+    return enumerate(fix_rows(read_rows(file_content)), start=2)
+
+
+def read_csv(file_content: bytes) -> Generator[dict[str, Any], None, None]:
+    try:
+        for row in csv.DictReader(StringIO(file_content.decode())):
+            yield {k.strip().lower(): v for k, v in row.items()}
```
rudigiesler, line 190: Are we not doing this in `fix_rows`? Maybe we can just return the DictReader here?

home/import_helpers.py
```diff
@@ -157,3 +163,55 @@ def extract_errors(data: dict[str | int, Any] | list[str]) -> dict[str, str]:
...
+
+    for row in worksheet.iter_rows(min_row=2, values_only=True):
+        r = {}
+        if len(row) > len(header):
+            raise ImportException(
+                "Invalid format. Please check that all row values have headers."
+            )
+        for name, cell in zip(header, row):  # noqa: B905
+            if name and cell:
+                r[name.strip().lower()] = clean_excel_cell(cell)
```
rudigiesler, line 215: Same here for the duplicate cleaning of keys. We'll probably still want the excel cell cleaning bit, because that's excel specific

home/import_assessments.py
```diff
@@ -385,16 +353,11 @@ def from_flat(cls, row: dict[str, str], row_num: int) -> "AssessmentRow":
...
-                key.strip(): value.strip()
-                for key, value in row.items()
-                if value and key.strip() in cls.fields()
-            }
-        except AttributeError:
-            raise ImportAssessmentException(
-                "Invalid format. Please check that all row values have headers."
-            )
+        row = {
+            key.strip(): value.strip()
```
rudigiesler, line 357: Do we need to strip again if the helper is doing this for us before we get the data?

home/import_helpers.py
HawkiesZA: I think this defends against a Nonetype exception. v.strip isn't possible if v is None.

HawkiesZA: Yup. Removing it and running the tests, you get 

`AttributeError: 'NoneType' object has no attribute 'strip'` on a few tests.

home/tests/test_assessment_import_export.py
```diff
@@ -530,21 +531,37 @@ def test_invalid_high_inflecton_csv_format(self, csv_impexp: ImportExport) -> No
...
-        with pytest.raises(ImportAssessmentException) as e:
+        with pytest.raises(ImportException) as e:
             csv_impexp.import_content_file("assessment_results.csv", purge=False)
             csv_impexp.import_file("extra_columns.csv")
-        assert (
-            e.value.message == "Invalid format. Please check that all row values "
-            "have headers."
-        )
+        assert e.value.message == [
+            "Invalid format. Please check that all row values " "have headers."
```
jerith, line 546:
```suggestion
            "Invalid format. Please check that all row values have headers."
```

home/import_helpers.py
rudigiesler: Yeah I'm thinking something more explicit would be better. Maybe we can separate it out into `normalise_key` and `normalise_value` functions, where we can have a
```
if v is None:
    return None
return v.strip()
```

But also, the typing for this function is incorrect, should the return then rather not be `dict[str, str | None]`?

home/import_helpers.py
HawkiesZA: read_csv and read_excel need to both return a Generator though

HawkiesZA: Looks like I can wrangle the types and turn read_csv into a one liner

home/import_assessments.py
```diff
@@ -385,16 +353,11 @@ def from_flat(cls, row: dict[str, str], row_num: int) -> "AssessmentRow":
...
+            for key, value in row.items()
+            if value and key.strip() in cls.fields()
```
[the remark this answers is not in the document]
HawkiesZA: This isn't new bahviour though, just removed the try/except.

home/tests/test_assessment_import_export.py
HawkiesZA: lol, what

home/import_helpers.py
```diff
@@ -165,16 +165,36 @@ def extract_errors(data: dict[str | int, Any] | list[str]) -> dict[str, str]:
     return error_message
 
 
-def fix_rows(rows: Generator[dict[str, str], None, None]) -> Iterator[dict[str, str]]:
+def fix_rows(
+    rows: Generator[dict[str, str], None, None] | Iterator[dict[str | Any, Any]]
```
jerith, line 169: Since we're not feeding anything into the generator, I'm pretty sure we can skip that type and just use `Iterator[dict[str | Any, Any]`. Although by the time we get here, we should be guaranteed string keys at least.

home/import_helpers.py
```diff
@@ -165,16 +165,36 @@ def extract_errors(data: dict[str | int, Any] | list[str]) -> dict[str, str]:
...
-    Fix keys for all rows by lowercasing and removing whitespace
+    Fix keys for all rows by lowercasing keys and removing whitespace from keys and values
     """
     for row in rows:
         yield fix_row(row)
 
 
-def fix_row(row: dict[str, str]) -> dict[str, str]:
-    return {k.lower().strip(): v.strip() if v else v for k, v in row.items()}
+def fix_row(row: dict[str, str]) -> dict[str, str | None]:
```
jerith, line 178: If `_normalise_value()` accepts `str | None`, `fix_row()` should accept `dict[str, str | None]`.

home/import_helpers.py
```diff
@@ -184,14 +204,8 @@ def parse_file(
...
-            yield {k.strip().lower(): v for k, v in row.items()}
-    except AttributeError:
-        raise ImportException(
-            "Invalid format. Please check that all row values have headers."
-        )
+def read_csv(file_content: bytes) -> Iterator[dict[str, str | None]]:
+    return csv.DictReader(StringIO(file_content.decode()))
 
 
 def read_xlsx(file_content: bytes) -> Generator[dict[str, Any], None, None]:
```
jerith, line 211: Can we make this `Iterator[dict[str, str | None]]` (or maybe `Any` for the dict's value type) to match `read_csv()`?

home/tests/test_assessment_import_export.py
```diff
@@ -530,21 +531,37 @@ def test_invalid_high_inflecton_csv_format(self, csv_impexp: ImportExport) -> No
...
+        column values than headers should return an intuitive error message
+
+        (This uses extra_columns.xlsx)
+        """
+
+        with pytest.raises(ImportException) as e:
+            csv_impexp.import_content_file("assessment_results.csv", purge=False)
+            xlsx_impexp.import_file("extra_columns.xlsx")
+        assert e.value.message == [
+            "Invalid format. Please check that all row values " "have headers."
```
jerith, line 563:
```suggestion
            "Invalid format. Please check that all row values have headers."
```

I'm guessing the formatter combined these lines or something.

home/import_helpers.py
```diff
@@ -165,21 +165,27 @@ def extract_errors(data: dict[str | int, Any] | list[str]) -> dict[str, str]:
...
     """
     try:
+        if index == 0:
+            keys = [_normalise_key(k) for k, v in row.items()]
+            if len(keys) != len(set(keys)):
+                raise ImportException(
+                    "Invalid format. Please check that there are no duplicate headers.",
+                    row_num=index,
+                )
+
```
jerith, line 188: Could we do this check up in `fix_rows()` with `if len(row) != len(fix_row(row))` on the first iteration? Perhaps with `first_row = next(rows)` before the loop?

HawkiesZA: Done

jerith (dismissed):

rudigiesler (dismissed):

rudigiesler (approved):

PR #398, opened 2024-12-04
Add go to form action types to buttons and lists

rudigiesler:
- **Add GoToForm button and list item types**
- **Add import/export for go to form buttons and list items**
- **Add go to form button and list items to API**

## Purpose
We currently have actions for viewing the next message, and going to a different content page. We also want to be able to start a form from a button or list item

## Solution
Added a new action type for buttons and lists, added it to the import and export for content pages, and added it to the content page API

#### Checklist
- [x] Added to release notes


home/tests/test_content_import_export.py
```diff
@@ -188,6 +195,22 @@ def _normalise_button_pks(body: DbDict, min_pk: int) -> DbDict:
                 button = button | {"value": v | {"page": v["page"] - min_pk}}
             buttons.append(button)
         value = value | {"buttons": buttons}
+    print("nomralised bttons ", body | {"value": value})
```
HawkiesZA, line 198: Can probably remove this :-)

rudigiesler: Haha, oh dear, my typo-ridden debug print forever immortalized in git history

HawkiesZA (approved):

home/import_content_pages.py
```diff
@@ -189,10 +190,11 @@ def add_go_to_page_buttons(self) -> None:
...
                             row.row_num,
                         )
-                    page.whatsapp_body[message_index].value["buttons"].append(
-                        ("go_to_page", {"page": related_page, "title": title})
+                    page.whatsapp_body[message_index].value["buttons"].insert(
+                        button["index"],
+                        ("go_to_page", {"page": related_page, "title": title}),
                     )
-            page.save_revision().publish()
+            page.save()
```
jerith, line 197: What implications does this change have? Will the updates to the buttons be unpublished?

rudigiesler: If I'm understanding the docs, it will update the currently published page. This is my understanding:
- The `Page` object in the database contains the current live/published content
- The `Revision` object contains the history of previous versions of the page
- So if you do a `save_revision`, it creates a new revision (a new point in the history). Because it's newer than the current page, it's seen as unpublished changes. If you then call `publish` on that, it copies whatever is in that revision onto the `Page`, making it live.

So if you just do a `save_revision`, then we'll have unpublished buttons until we call `publish` on that revision. If we do a `save`, it should update the published page with new buttons.

I think that's okay, unless there's something I'm not thinking about?

jerith: If you're happy with that, I am too -- I've never been able to wrap my head around wagtail's behaviour. Do we support importing unpublished content? If so, do we have tests for the behaviour there?

jerith (approved):

home/import_content_pages.py
rudigiesler: No we don't support that. Export and import only works on published content, we assume that anything you're importing you want published when the import is done.

We also don't import/export revision history, so by doing a "purge" import you do lose that history (there's also things like comments and moderation review that we don't import/export).

PR #401, opened 2024-12-17
Form answer responses

rudigiesler:
## Purpose
For each answer to a form, we want an optional response field.

This field will be used in things like quizzes, where we want to give immediate feedback to a user depending on which option they chose, e.g. If they chose the wrong answer, we want to tell them that it is incorrect, what the correct answer is, and why that is the correct answer.

## Solution
Added the field to the model definition. Added to import/export. Ensured that it was present in the API.

For the import/export tests, I also changed from a mix of generator and list returns to just list returns. The reasoning here is that there was a bug caused by the mixing of the types (one of the methods called something like `max` on the iterator, and then tried to iterate through the now empty iterator). I chose lists over generators because we often needed to go through the whole list twice, so we would be duplicating in memory with `tee` anyway.

#### Checklist
- [ ] Added to release notes


home/tests/test_api.py
```diff
@@ -1924,8 +1924,18 @@ def create_test_data(self):
...
+                    "score": "5",
+                    "semantic_id": "crunchie",
+                    "response": "Get that Friday feeling",
+                },
+                {
+                    "answer": "Flake",
+                    "score": "3",
+                    "semantic_id": "flake",
+                    "response": "Only the flakiest",
+                },
```
jerith, line 1938: Since `response` is optional, do we need a test where it's empty?

home/tests/test_assessment_import_export.py
```diff
@@ -79,16 +79,17 @@ def decode_json_fields(page: DbDict) -> DbDict:
 
 def _remove_fields(pages: DbDicts, field_names: set[str]) -> DbDicts:
     for p in pages:
-        fields = {k: v for k, v in p["fields"].items() if k not in field_names}
-        yield p | {"fields": fields}
+        p["fields"] = {k: v for k, v in p["fields"].items() if k not in field_names}
+    return pages
```
jerith, line 83: This change modifies the input, which the previous implementation didn't do. Is that guaranteed safe? Should we perhaps build a new list instead modifying the items in the one we get?

(Several other changes in this file do the same -- I won't note each case individually for now.)

rudigiesler: Yes I did think of this, but the way we're using it is `assessments = ff(assessments)`, so we're throwing away the old list on every application of a filter anyway, so I thought it would be fine to modify in place.

jerith (approved):

PR #407, opened 2025-01-28
Add language column in import export

Hlamallama:
## Purpose
_Why does this exist?_

## Solution

#### Checklist
- [ ] Added to release notes


home/export_content_pages.py
```diff
@@ -80,6 +81,7 @@ class ExportRow:
     media_link: str = ""
     related_pages: str = ""
     footer: str = ""
+    language_code: str = ""
```
rudigiesler, line 84: I think we should also remove the `locale` from the export, so that we have just a single column referencing the language.

home/export_content_pages.py
```diff
@@ -208,6 +210,11 @@ def _export_content_page(self, page: ContentPage, structure: str) -> None:
          * We should use the parent slug (which is expected to be unique per
            locale (probably?)) instead of the parent title.
         """
+        export_lang_code = ""
+        for lang_code, lang_dn in get_content_languages().items():
+            if lang_dn == str(page.locale):
+                export_lang_code = lang_code
```
rudigiesler, line 216: Can we not just do `page.locale.language_code` to get the language code for the current page?

home/import_content_pages.py
```diff
@@ -268,7 +268,10 @@ def create_content_page_index_from_row(self, row: "ContentRow") -> None:
     def create_shadow_content_page_from_row(
         self, row: "ContentRow", row_num: int
     ) -> None:
-        locale = self.locale_from_display_name(row.locale)
+        if row.language_code:
+            locale = locale = Locale.objects.get(language_code=row.language_code)
```
rudigiesler, line 272: What if no Locale exists for the language code given in the row? In that case, we should raise an `ImportException`, similar to how `locale_from_display_name` does.

home/tests/import-export-data/language_code_import.csv
```diff
@@ -0,0 +1,35 @@
...
+_Stop smoking with the help of a guided, 42-day program._
+2. Manage your stress 🧘🏽‍♀️
+_Learn how to cope with stress and improve your wellbeing._",,UTILITY,[],,,,[],,,,,self-help,"*Self-help programs* 🌬️
+
+Reply with a number to take part in a *free* self-help program created by WHO.
+
+1. Quit tobacco 🚭
+_Stop smoking with the help of a guided, 42-day program._
+2. Manage your stress 🧘🏽‍♀️
+_Learn how to cope with stress and improve your wellbeing._",,,3e5d77f7-4d34-430d-aad7-d9ca01f79732,self_help,,,English,,[],,,,,,
```
rudigiesler, line 35: Do we need so much and so long content for this test? Maybe we can reduce it to just what is necessary, so that it's easier to see what we're testing here.

home/export_content_pages.py
```diff
@@ -208,6 +210,11 @@ def _export_content_page(self, page: ContentPage, structure: str) -> None:
          * We should use the parent slug (which is expected to be unique per
            locale (probably?)) instead of the parent title.
         """
+        export_lang_code = ""
+        for lang_code, lang_dn in get_content_languages().items():
```
rudigiesler, line 214: It's not just content pages that need language codes, it's also content index pages.

home/export_content_pages.py
```diff
@@ -208,6 +210,11 @@ def _export_content_page(self, page: ContentPage, structure: str) -> None:
          * We should use the parent slug (which is expected to be unique per
            locale (probably?)) instead of the parent title.
```
rudigiesler, line 211: We can update this FIXME now that we're using the language code.

home/import_content_pages.py
```diff
@@ -252,7 +259,7 @@ def create_content_page_index_from_row(self, row: "ContentRow") -> None:
         # but optional for the default locale.
         if row.translation_tag or locale != self.default_locale():
             index.translation_key = row.translation_tag
-        locale = self.locale_from_display_name(row.locale)
+        # locale = self.locale_from_display_name(row.locale)
```
rudigiesler, line 262: We can remove this line instead of commenting it out

rudigiesler (commented):
Looks good! Just one tiny comment, and then could we add this to CHANGELOG.md in this PR?

Then we'll also need to update the docs. Those are in the wiki though, so not something we can do in this PR, so will have to be done after merging this.

rudigiesler (approved):

PR #408, opened 2025-02-05
Assessment import for multiple languages and identical slug

DevChima:
## Purpose
_When looking up content pages in the assessments import, we're only doing so from the slug, not the locale. So if there's a page with a translation, we'll find multiple results and the import will break._

## Solution
_Look up content pages with both slug and locale_

#### Checklist
- [ ] Added to release notes


home/import_assessments.py
```diff
@@ -398,15 +406,19 @@ def from_flat(cls, row: dict[str, str], row_num: int) -> "AssessmentRow":
...
         raise ImportAssessmentException(
             f"You are trying to add an assessment, where one of the result pages "
-            f"references the content page with slug {slug} which does not exist. "
+            f"references the content page with slug {slug} and locale {locale} which does not exist. "
             "Please create the content page first."
         )
+    except ContentPage.MultipleObjectsReturned:
+        raise ImportAssessmentException(
+            f"Multiple ContentPages found with slug '{slug}' and locale '{locale}'. Please ensure slugs are unique within each locale."
+        )
```
rudigiesler, line 421: There's no test for this, but also it should be possible to have the same slug for two pages in the same locale, so I don't think we can create a test for this? If we can't create a test because it's not possible to get in this situation, then we can probably remove this?

home/import_assessments.py
```diff
@@ -398,15 +406,19 @@ def from_flat(cls, row: dict[str, str], row_num: int) -> "AssessmentRow":
             )
 
 
-def get_content_page_id_from_slug(slug: str) -> int:
+def get_content_page_id_from_slug(slug: str, locale: str) -> int:
```
rudigiesler, line 409: Instead of sending the language code here, and then doing a join in the db, why not just send the actual locale object to this function? i.e. `locale: Locale`

Then calling the function becomes `get_content_page_id_from_slug(self.low_result_page, self.locale)`, and the filter becomes `ContentPage.objects.get(slug=slug, locale=locale)`, simplifying it.

rudigiesler:
Also, could you update the release notes?

home/import_assessments.py
DevChima: Yes, we can't create a test for this with this new fix so I'll remove it.

rudigiesler (approved):

PR #409, opened 2025-02-05
remove hidden characters in import files

Hlamallama:
## Purpose
_Why does this exist?_

## Solution

#### Checklist
- [ ] Added to release notes

### DIFF
--- a/home/import_content_pages.py
+++ b/home/import_content_pages.py
@@ -6,7 +6,10 @@ from dataclasses import dataclass, field, fields
 from json.decoder import JSONDecodeError
 from queue import Queue
 from typing import Any
+import unicodedata
 from uuid import uuid4
+import re
+from dataclasses import replace
 
 from django.core.exceptions import ObjectDoesNotExist, ValidationError  # type: ignore
 from taggit.models import Tag  # type: ignore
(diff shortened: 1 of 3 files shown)


home/tests/import-export-data/content2.csv
```diff
@@ -24,12 +24,12 @@ Reply with a number to take part in a *free* self-help program created by WHO.
 
 1. Quit tobacco 🚭
 _Stop smoking with the help of a guided, 42-day program._
-2. Manage your stress 🧘🏽‍♀️
+2. Manage your stress
```
jerith, line 27: I think `content2.csv` is used in several places, so let's rather add a modified copy (to use for whichever test this is for) instead of modifying `content2.csv` itself.

home/tests/test_content_import_export.py
```diff
@@ -1665,6 +1665,12 @@ def test_language_code_import(self, csv_impexp: ImportExport) -> None:
         assert dst == src
 
 
+    def test_hidden_characters(self, csv_impexp: ImportExport) -> None:
+        """ """
```
jerith, line 1669: This test needs a proper docstring.

jerith (commented):
I've made a couple of comments, but I think someone who understands wagtail validation better than I do needs to review that part.

home/models.py
```diff
@@ -973,6 +977,20 @@ def clean(self):
         result = super().clean(Page)
         errors = {}
 
+        # Clean the whatsapp body to remove hidden characters
+        if self.whatsapp_body and isinstance(self.whatsapp_body, StreamValue):
+            whatsapp_block = []
```
rudigiesler, line 982: We're not using this variable anywhere, so we can remove it

home/tests/test_content_import_export.py
```diff
@@ -576,7 +576,7 @@ def test_simple(self, csv_impexp: ImportExport) -> None:
          * Should we set enable_web and friends based on body, title, or an
            enable field that we'll need to add to the export?
         """
-        csv_bytes = csv_impexp.import_file("content2.csv")
+        csv_bytes = csv_impexp.import_file("content3.csv")
```
rudigiesler, line 579: Why are we changing to `content3.csv` here?

home/tests/test_content_import_export.py
```diff
@@ -1664,6 +1666,13 @@ def test_language_code_import(self, csv_impexp: ImportExport) -> None:
         src, dst = csv_impexp.csvs2dicts(new_export_content, content)
         assert dst == src
 
+    def test_hidden_characters(self, csv_impexp: ImportExport) -> None:
+        """
+        Import a page that has hidden characters in the whatsapp body
+        """
+        csv_impexp.import_file("test_special_chars.csv")
```
rudigiesler, line 1673: Could we assert that the character is present in this source file?

home/import_content_pages.py
```diff
@@ -8,7 +8,10 @@
 from typing import Any
 from uuid import uuid4
 
-from django.core.exceptions import ObjectDoesNotExist, ValidationError  # type: ignore
+from django.core.exceptions import (
+    ObjectDoesNotExist,  # type: ignore
+    ValidationError,
```
rudigiesler, line 13: I don't think this should've changed, it's causing a typing error in the linting now

rudigiesler (commented):
Looks good! I think that there are some changes to the formatting of the imports on some files that is unnecessary, and causing the linting to fail. Calling `ruff` on them should put them back.

CHANGELOG.md
```diff
@@ -10,9 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Answer responses for Forms
 - Language_code field on imports and exports
+- whatsapp body hidden characters validation
```
rudigiesler, line 13: I wouldn't call this validation, I think "removal" is a better word here, since we're automatically removing them from the whatsapp body

home/tests/import-export-data/content2.csv
rudigiesler: So what we're doing here is just removing the non-printable characters from this CSV, because otherwise the import-export-compare tests fail because those characters have been removed. I've said that it's okay to modify this CSV in this way, since we have tests specific to the removal of chars, so we don't need those chars also in this CSV.

rudigiesler (commented):
Just a small comment on the wording of the changelog, otherwise all looks good!

rudigiesler (approved):

PR #412, opened 2025-02-12
add error message for failing list items

Hlamallama:
## Purpose
_Why does this exist?_

## Solution

#### Checklist
- [x] Added to release notes


home/import_content_pages.py
```diff
@@ -418,8 +418,8 @@ def add_message_to_shadow_content_page_from_row(
                         )
                 except ImportException:
                     raise
-                except Exception:
-                    raise ImportException("Invalid list_items")
+                except Exception as e:
+                    raise ImportException(f"List item is missing key {e}")
```
jerith, line 422: What if the exception isn't a `KeyError`?

home/tests/test_content_import_export.py
```diff
@@ -1674,6 +1674,17 @@ def test_hidden_characters(self, csv_impexp: ImportExport) -> None:
         assert "\u2028\u2028" in csv_bytes.decode("utf-8")
         assert "\u2028" not in ContentPage.objects.all().values()[0]["whatsapp_body"]
 
+    def test_list_item_descriptive_error_message(
+        self, csv_impexp: ImportExport
+    ) -> None:
+        """
+        Import failure for list_items should return a descriptive error message
+        """
```
jerith, line 1682: Can we try add a test for a different kind of error to make sure that code path works as expected?

home/import_content_pages.py
```diff
@@ -385,34 +385,48 @@ def add_message_to_shadow_content_page_from_row(
...
-                            "id": str(uuid4()),
-                            "type": item["type"],
-                            "value": {"title": item["title"], "form": form.id},
-                        }
-                    )
+                except ImportException:
+                    raise
+                except KeyError as e:
+                    raise ImportException(f"List item is missing key {e}")
+                except Exception as e:
```
rudigiesler, line 427: I don't think we should be catching just a generic `Exception` here, since we can't create a good message to tell the user what is wrong. We also don't have any tests for it, which shows that we don't really know what issues we're trying to catch here. If there are other known issues, then we should be catching those specifically with a proper error message and tests. If not, then I would just suggest just removing this block

Hlamallama: Currently I don't know any other specific issue that we can catch here, I'll remove the generic exception

erikh360 (approved):

PR #417, opened 2025-02-17
refactor add_message_to_shadow_content_page_from_row

Hlamallama:
## Purpose
_Why does this exist?_

## Solution

#### Checklist
- [ ] Added to release notes


home/import_content_pages.py
```diff
@@ -339,91 +342,110 @@ def add_variation_to_shadow_content_page_from_row(
...
+    def _create_interactive_items(
+        self,
+        row_field: list[dict[str, Any]],
+        page: Page,
+        slug: str,
+        locale: Locale,
+        field_name: str,
+    ) -> list[dict[str, Any]]:
+        messages: list[dict[str, Any]] = []
+        for index, content_page_field in enumerate(row_field):
```
jerith, line 373: These aren't messages or fields, so maybe:
- `item_type` for the parameter (which is either `"button"` or `"list item"`)
- `items` for the list we're collecting into (instead of `messages`)
- `item` for the dict we're processing (instead of `field`)

would be better names?

home/import_content_pages.py
```diff
@@ -60,6 +60,9 @@ def __init__(
         self.go_to_page_list_items: dict[PageId, dict[int, list[dict[str, Any]]]] = (
             defaultdict(lambda: defaultdict(list))
         )
+        self.go_to_page: dict[PageId, dict[int, list[dict[str, Any]]]] = defaultdict(
+            lambda: defaultdict(list)
+        )
```
jerith, line 65: Is this used anywhere?

jerith (commented):
The second part of this work is to combine the `add_go_to_page_buttons` and `add_go_to_page_list_items` methods similar to what we've just done with with the duplicated code in `add_message_to_shadow_content_page_from_row`.

home/import_content_pages.py
```diff
@@ -216,11 +191,11 @@ def add_go_to_page_list_items(self) -> None:
                         row = self.shadow_pages[(slug, locale)]
                         raise ImportException(
                             f"No pages found with slug '{item['slug']}' and locale "
-                            f"'{locale}' for go_to_page list item '{item['title']}' on "
+                            f"'{locale}' for go_to_page {item_type[:-1]} '{item['title']}' on "
                             f"page '{slug}'",
                             row.row_num,
                         )
-                    page.whatsapp_body[message_index].value["list_items"].insert(
+                    page.whatsapp_body[message_index].value[item_type].insert(
```
jerith, line 198: If this is the only place we care about whether we have a button or a list item, we can probably include this key in the data structure we put in `items_dict` and consolidate things even further. However, I've spent over an hour trying to figure out how all the button/list stuff fits together so I can start looking into this consolidation and not had much success, so I think let's leave it as-is for now and try fix it separately.

home/import_content_pages.py
```diff
@@ -339,91 +317,108 @@ def add_variation_to_shadow_content_page_from_row(
...
+                    items.append(
                         {
                             "id": str(uuid4()),
-                            "type": button["type"],
-                            "value": {"title": button["title"], "form": form.id},
+                            "type": item["type"],
+                            "value": {
+                                "title": item["title"],
+                                "form": form.id,
+                            },
```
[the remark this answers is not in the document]
Hlamallama: black doesn't like it when it is one line

jerith (approved):

PR #419, opened 2025-02-17
CMS Forms Flexible Imports

DevChima:
## Purpose
_There is need for CMS forms to have a more robust import mechanism._
_Also, we need the error handling to be more intuitive so there's better guidance for the user when there's an import failure_

## Solution
_This pr includes:_
_A header validator function which validates the header before writing to rows_
_A list of mandatory headers that each import should have_
_A snakecase function which converts the headers to snake case so the import is more flexible_
_What this means is that if a user enters Generic Error instead of generic_error as a header name, it will convert and pass._

#### Checklist
- [ ] Added to release notes


home/tests/test_assessment_import_export.py
```diff
@@ -425,6 +425,41 @@ def test_single_assessment(self, impexp: ImportExport) -> None:
...
+        (This uses snake_case_assessments.csv.)
+        """
+        csv_impexp.import_file("snake_case_assessments.csv")
+        imported_assessments = Assessment.objects.all()
+        assert imported_assessments.exists()
+        assessment_model = Assessment
+        model_fields = [field.name for field in assessment_model._meta.get_fields()]
+
+        expected_fields = {
+            "title": True,
```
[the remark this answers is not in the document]
DevChima: Actually that's true.

home/import_assessments.py
```diff
@@ -96,9 +98,32 @@ def save_assessments(self) -> None:
...
+        If there are rows:
+            a. Extracts the original headers from the first row.
+            b. Converts the original headers to snake_case and creates a mapping from original headers to snake_case headers.
+            c. Validates that the snake_case headers contain all mandatory headers.
+            d. Transforms each row to use snake_case headers.
+        """
+        row_iterator = parse_file(self.file_content, self.file_type)
+        rows = [row for _, row in row_iterator]
+
+        if rows:
```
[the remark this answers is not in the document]
DevChima: I've added it in but I also had to modify the helper function here:
https://github.com/praekeltfoundation/contentrepo/blob/ad6ca7bbbd730b33c49518ea4114b66388847187/home/import_helpers.py#L172

It would throw a runtime stop-iteration error becuase it will try use next(rows) on an empty iterator since the file is empty. 
So I put it in a try-except block to catch the exception if the iterator is empty.
Then in the parse_file function, I added an if not rows to catch emoty rows. 

@erikh360

home/import_assessments.py
```diff
@@ -185,6 +214,21 @@ def create_shadow_assessment_from_row(
         )
         assessment.questions.append(question)
 
+    def validate_headers(
```
erikh360, line 217: Do we have to pass in MANDATORY_HEADERS here? validate_headers is in the same class where it is called and can also access the variable directly.

home/import_assessments.py
```diff
@@ -96,9 +98,36 @@ def save_assessments(self) -> None:
...
+
+        original_headers = rows[0].keys()
+        headers_mapping = {
+            header: self.to_snake_case(header) for header in original_headers
+        }
+        snake_case_headers = list(headers_mapping.values())
+        self.validate_headers(snake_case_headers, MANDATORY_HEADERS, row_num=1)
+        transformed_rows = [
+            {headers_mapping[key]: value for key, value in row.items()} for row in rows
+        ]
```
erikh360, line 126: Should some of this not maybe live inside the parse_file function? I see we have a function in there called fix_rows?

I'm not very familiar with all the import code so I'm just asking

home/tests/import-export-data/broken_assessment.csv
```diff
@@ -1,2 +1,2 @@
 this,is,not,a,valid,content,csv
-"For real, it's totally not."
+"For real, it's totally not.",,,,,,
```
jerith, line 2: The point of this file is to be a broken csv -- adding the commas makes it not broken anymore. 😅

home/tests/test_assessment_import_export.py
```diff
@@ -618,6 +655,64 @@ def test_invalid_medium_score(self, csv_impexp: ImportExport) -> None:
...
+        Importing a CSV with a missing generic error field should return an error
+        that a generic error is mmissing
+        """
+        with pytest.raises(ImportAssessmentException) as e:
+            csv_impexp.import_file("assessments_missing_generic_error.csv")
+        assert (
+            e.value.message
+            == "The import file is missing required fields: generic_error"
+        )
+        assert e.value.row_num == 2
```
jerith, line 702: We probably only need one of these, since we test the "all columns missing" case above.

home/import_helpers.py
```diff
@@ -165,11 +166,39 @@ def extract_errors(data: dict[str | int, Any] | list[str]) -> dict[str, str]:
...
+    Converts a list of headers to snake_case and returns a mapping.
+    """
+    return {header: to_snake_case(header) for header in headers}
+
+
+def to_snake_case(s: str) -> str:
+    """
+    Converts string to snake_case.
+    """
+    return re.sub(r"[\W_]+", "_", s).lower().strip("_")
```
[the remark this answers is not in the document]
DevChima: If we do it that way then making it reusable across all imports could be more cumbersome since they all have their various headers?

home/import_assessments.py
```diff
@@ -96,9 +104,31 @@ def save_assessments(self) -> None:
             )
 
     def parse_file(self) -> list["AssessmentRow"]:
+        """
+        Converts the iterator to a list of rows.
+        If there are rows:
+            a. Extracts the original headers from the first row.
+            b. Converts the original headers to snake_case and creates a mapping from original headers to snake_case headers.
+            c. Validates that the snake_case headers contain all mandatory headers.
+            d. Transforms each row to use snake_case headers.
```
jerith, line 113: The snake_case transformation probably belongs in the shared code, so all importers get cleaned-up keys. That would also avoid having to juggle the mapping from old to new keys in here.


Can we do that without breaking anything in the other importers?

home/import_assessments.py
```diff
@@ -360,6 +400,18 @@ class AssessmentRow:
...
+        """
+        Checks for missing required fields in the row and raises an exception if any is missing.
+        """
+        missing_fields = [field for field in MANDATORY_HEADERS if field not in row]
+        if missing_fields:
+            raise ImportAssessmentException(
+                f"The import file is missing required fields: {', '.join(missing_fields)}",
+                row_num,
+            )
+
```
jerith, line 414: Isn't this already covered by `AssessmentImporter.validate_headers`?

home/import_assessments.py
DevChima: The snake_case even though it's in helpers.py is not shared otherwise it will break contentset importer which looks up headers in pascal case. if we needed to integrate it in the generic import process then we would first need to fix the code in content pages.

home/import_assessments.py
DevChima: No, validate_headers checks for missing headers before writing to rows. Check_missing_fields then checks that the mandatory headers do not have missing fields in its rows.

jerith: All rows will have exactly the same set of keys, so if the headers aren't missing there they won't be missing here. If the intent is to check for missing _values_, we should change the code to do that instead of just checking the keys.

(Unless we're removing empty fields somewhere along the way and I missed it?)

home/import_assessments.py
jerith: Ah. In that case, can you add a `TODO` comment to move this to the shared code once we're able to work on the contentpage import without interfering with other ongoing work?

home/import_assessments.py
```diff
@@ -394,8 +433,14 @@ def from_flat(cls, row: dict[str, str], row_num: int) -> "AssessmentRow":
...
             )
         except TypeError:
+            missing_fields = [
+                field
+                for field in MANDATORY_HEADERS
+                if field not in row or row[field] == ""
+            ]
             raise ImportAssessmentException(
-                "The import file is missing some required fields."
+                f"The import file is missing required fields: {', '.join(missing_fields)}",
```
jerith, line 442: Perhaps `"Row missing values for required fields: {...}"` instead?

jerith (approved):

PR #427, opened 2025-02-24
added missing ordered set page slug validation

Mudiwa66:
## Purpose
When you import an ordered content set with EDD, Unit etc, but without the page slug, then the imported set will not have that data

## Solution
The CMS does not allow imports for ordered sets with page details but no page linked

#### Checklist
- [X] Added or updated unit tests
- [ ] Added to release notes

### DIFF
--- a/home/import_ordered_content_sets.py
+++ b/home/import_ordered_content_sets.py
@@ -162,7 +162,7 @@ class OrderedContentSetImporter:
                 page.time or page.unit or page.before_or_after or page.contact_field
             ):
                 raise ImportException(
-                    f"You are attempting to import an ordered content set with page details, but no page slug.",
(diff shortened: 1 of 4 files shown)


home/import_ordered_content_sets.py
```diff
@@ -158,6 +158,13 @@ def _add_pages(
                         f"Content page not found for slug '{page.page_slug}' in locale '{locale}'",
                         index,
                     )
+            if not page.page_slug and (
+                page.time or page.unit or page.before_or_after or page.contact_field
+            ):
```
jerith, line 163: Looking at the condition above, we appear to have a gap where `page_slug` is `"-"` or `page_slug` is empty and we don't also have one of these other fields. What should happen in these cases?

HawkiesZA (approved):

PR #429, opened 2025-02-26
Cms forms fix xls import

DevChima:
## Purpose
_CMS forms fails to import when the file is an XLS and some headers have missing values._
_This happens because the read_xlsx helper function skips headers in rows where the corresponding cell is empty which then causes a mismatch that leads to a keyerror when we try to get keys here: header_mapping[key]._
_This pr does not try to modify read_xlsx() since it's also used by content_pages but instead we modify the parse_file_
_transformation to accomodate for missing values._.

## Solution
_Ensure that we only map keys that exist in headers_mapping by using headers_mapping.get(key, key)_
_NB: Down the line, AssessmentRow.from_flat handles missing fields with defaults_

#### Checklist
- [x] Added to release notes


jerith (commented):
Since all our importers assume that they get the same data from both CSV and XLSX, this fix belongs in `read_xlsx` so all importers can benefit from it.

home/import_helpers.py
```diff
@@ -262,24 +262,33 @@ def read_csv(file_content: bytes) -> Iterator[dict[str, Any]]:
...
 
-    def clean_excel_cell(cell_value: str | float | datetime | None) -> str:
-        return str(cell_value).replace("_x000D", "").strip()
-
     first_row = next(worksheet.iter_rows(max_row=1, values_only=True))
     header = [clean_excel_cell(cell) if cell else None for cell in first_row]
+    header = remove_trailing_nones(header)
 
     for row in worksheet.iter_rows(min_row=2, values_only=True):
+        row = remove_trailing_nones(row)  # type: ignore
```
jerith, line 284: Can you add a comment explaining why we're ignoring the type here?

jerith (approved):

PR #434, opened 2025-03-04
Re-instate import_json_content_turn.py and remove import_json_content.py

DevChima:
## Purpose
_Removes management command import_json_content.py and returns import_json_content_turn.py which was earlier removed_

## Solution
_In a pr from yesterday we removed import_json_content_turn.py (name mis-match) but we need it so we're returning it and removing  import_json_content.py management command instead. It didn't have any tests so tests will be added in a seperate pr_

#### Checklist
- [ ] Added to release notes


home/management/commands/import_json_content_turn.py
```diff
@@ -0,0 +1,94 @@
...
+                    if translations:
+                        translation = translations.first()
+                        if translation.pk != contentpage.pk:
+                            key = translation.translation_key
+                            contentpage.translation_key = key
+                            contentpage.save_revision().publish()
+                except Exception as e:
+                    print(e)
+
+            self.stdout.write(self.style.SUCCESS("Successfully imported Content Pages"))
```
rudigiesler, line 94: If we're keeping this, then we should have tests for it, to ensure that it actually works correctly.

DevChima: So the plan was to revert that PR which removed it and then in a seperate pr, push tests for it. 
I can go ahead and add the tests in this pr if that's the case.

rudigiesler (approved):

PR #439, opened 2025-03-31
Remove quick replies from standalone templates and replace with real buttons

HawkiesZA:
## Purpose
Quick replies don't keep their ordering, nor do they allow us to apply different click types, which buttons allow us to do.

## Solution
We strip out quick replies from WA Templates to use buttons instead.

#### Checklist
- [x] Added to release notes


home/tests/import-export-data/whatsapp-template-simple.csv
```diff
@@ -1,2 +1,2 @@
-name,category,quick_replies,locale,image,message,example_values,submission_name,submission_status,submission_result
-Test Template Simple,MARKETING,,en,,Message,,,,
+name,category,buttons,locale,image,message,example_values,submission_name,submission_status,submission_result
+Test Template Simple,MARKETING,"[{""type"": ""next_message"", ""title"": ""Sure, remind me"", ""slug"": """"}, {""type"": ""next_message"", ""title"": ""No, go to main menu"", ""slug"": """"}]",en,,Message,,,,
```
rudigiesler, line 2: Maybe one of these buttons should be a `go_to_page`, and one a `go_to_form`, so that we can test all the button types for import?

home/tests/test_models.py
```diff
@@ -1043,11 +1054,6 @@ def test_template_create_with_buttons(self):
             locale=Locale.objects.get(language_code="en"),
         )
         wat.save()
-        created_qr, _ = TemplateContentQuickReply.objects.get_or_create(name="Button1")
-        wat.quick_replies.add(created_qr)
-        created_qr, _ = TemplateContentQuickReply.objects.get_or_create(name="Button2")
-        wat.quick_replies.add(created_qr)
-        wat.save()
```
rudigiesler, line 1050: I think we'll need to add these into the `WhatsAppTemplate` as buttons, otherwise the `test_template_create_with_buttons` test name no longer makes sense..

home/import_whatsapp_templates.py
```diff
@@ -141,13 +152,78 @@ def default_locale(self) -> Locale:
...
+                elif item["type"] == "go_to_page":
+                    item["index"] = index
+                    if item_type == "button":
+                        go_to_page = self.go_to_page_buttons
+                    else:
+                        raise ImportException(
+                            f"{item_type} with invalid type '{item['type']}'"
+                        )
+                    page_gtp = go_to_page[(template_name, locale)]
+                    page_gtp.append(item)
```
rudigiesler, line 182: I would think that we would need to add something to `items`? But maybe updating the tests to create all the button types will surface this.

HawkiesZA: Forgot a step. When importing content pages, we build up the go_to_page buttons, and then later add them to the model. Forgot that second step.

rudigiesler (approved):

PR #441, opened 2025-04-07
Link templates to pages

HawkiesZA:
## Purpose
At the moment, templates are specified by setting certain fields inside a content page.

We want to change this so that the standalone templates (separate model) are used instead, and instead you can create a link from the content page to the whatsapp template.

## Solution
This PR allows pages to link to standalone templates.

#### Checklist
- [x] Added to release notes


contentrepo/settings/base.py
```diff
@@ -266,5 +266,5 @@
 
 # Flag for turning on Standalone Whatsapp Templates, still in development
 ENABLE_STANDALONE_WHATSAPP_TEMPLATES = env.bool(
-    "ENABLE_STANDALONE_WHATSAPP_TEMPLATES", False
+    "ENABLE_STANDALONE_WHATSAPP_TEMPLATES", True
```
rudigiesler, line 269: I think we should either keep this default to False, or if it breaks things now if set to False, then we should remove it

home/models.py
```diff
@@ -1004,12 +1014,16 @@ def tag(self):
...
-    def wa_body(self):
+    def wa_body(self) -> str:
         body = (
-            "\n".join(m.value["message"] for m in self.whatsapp_body)
+            "\n".join(
+                m.value["message"] if is_subscriptable(m.value) else ""
+                for m in self.whatsapp_body
+            )
             if self.whatsapp_body
             else ""
```
rudigiesler, line 1024: If it's a template, do we instead want to show the template text here instead? I think this is used in list views of content pages.

home/models.py
```diff
@@ -1004,12 +1014,16 @@ def tag(self):
...
         )
+        print(f"BODY: {body}")
```
rudigiesler, line 1026: Can we remove this?

home/tests/test_whatsapp_template_import_export.py
```diff
@@ -249,7 +247,7 @@ def xlsx_impexp(request: Any, admin_client: Any) -> ImportExport:
 
 
 # Needs this here while standalone templates are hidden behind feature flag
-register_snippet(WhatsAppTemplateAdmin)
+# register_snippet(WhatsAppTemplateAdmin)
```
rudigiesler, line 250: We can remove this completely instead of commenting it out

home/models.py
```diff
@@ -594,6 +598,12 @@ class WhatsAppTemplateCategory(models.TextChoices):
                     help_text="Each message will be sent with the text and media"
                 ),
             ),
+            (
+                "Whatsapp_Template",
+                SnippetChooserBlock(
+                    "home.WhatsAppTemplate", help_text="WhatsAppTemplate to use"
+                ),
+            ),
```
rudigiesler, line 606: I think we should add a test at least in `test_models` to ensure that we can save a link between the two, and then assert that it can display correctly (e.g. `wa_body`)

rudigiesler (commented):

rudigiesler (approved):

rudigiesler (approved):

PR #442, opened 2025-04-09
Cleanup whatsapp templates

fritzbrand:
## Purpose
Before taking the next steps on the Standalone Template work, we need to clean up the bits we left hanging when we paused this work

## Solution
This PR cleans up a lot of the old code, splits the standalone tests into their own file, and adds some better validation and tests

#### Checklist
- [ ] Added to release notes


home/models.py
```diff
@@ -1830,30 +1852,29 @@ def clean(self):
...
         for ev in example_values:
             if "," in ev["value"]:
                 errors["example_values"] = ValidationError(
                     "Example values cannot contain commas"
                 )
 
-        if mismatches:
+        message = self.message
+        # Matches "{1}" and "{11}", not "{{1}", "{a}" or "{1 "
+        single_braces = re.findall(r"[^{]{(\d*?)}", message)
```
rudigiesler, line 1864: I think this might not work where the variable is the first thing in the message, e.g. `{1}, it's time for your clinic appointment`

home/models.py
```diff
@@ -1830,30 +1852,29 @@ def clean(self):
...
             errors.setdefault("message", []).append(
                 ValidationError(
-                    f"Please provide variables with matching braces. You provided {mismatches}."
+                    f"Please provide variables with valid double braces. You provided single braces {single_braces}."
                 )
             )
 
+        brace_mismatches = self.check_matching_braces(message)
+
+        if brace_mismatches:
```
rudigiesler, line 1875: These if statements might be a good candidate for the new walrus operator, e.g. `if brace_mismatches := self.check_matching_braces(message):`

home/tests/test_models.py
```diff
@@ -1056,6 +1065,7 @@ def test_simple_template_submission(self) -> None:
         }
 
     # TODO: Find a better way to test quick replies
+    # Fritz -> I'm not sure what I meant here, nor do I have any ideas.  Will ask Rudi and/or Jeremy
```
rudigiesler, line 1068: We're doing away with quick replies and replacing them with buttons, so this is perhaps not as necessary. (depending on how the buttons are tested)

home/whatsapp.py
```diff
@@ -199,14 +198,25 @@ def submit_whatsapp_template(
     if response.ok:
         return response.json()
     else:
-        # TODO: Add better error handling to differentiate between user error or server error
-        raise TemplateSubmissionException(response.json())
+        if response.status_code >= 500:
+            raise TemplateSubmissionServerException(str(response.content))
+        if 400 <= response.status_code < 500:
+            if "application/json" in response.headers.get("Content-Type", ""):
+                raise TemplateSubmissionClientException(response.json())
```
rudigiesler, line 205: If the headers are wrong, then we would still raise an unhandled exception here. Also, if the headers say it's not JSON, then we don't raise any error, which seems like the incorrect things to do here.

Maybe we can do:
```python
try:
  raise TemplateSubmissionClientException(response.json())
except JSONDecodeError:
  raise TemplateSubmissionClientException({"error": {"error_user_msg": str(response.content)}})
  ```
  
 Or even better, if we remove the error message extraction from the exception, then we can do:
 ```python
try:
  raise TemplateSubmissionClientException(response.json()["error"]["error_user_msg"])
except JSONDecodeError, KeyError, TypeError:
  raise TemplateSubmissionClientException(str(response.content))
  ```
  
  We probably also want tests for all these cases (if they don't already exist and I missed them), to ensure that they're all saving errors correctly.

File: pyproject.toml [hunk 5]
```diff
@@ -128,7 +128,9 @@ filterwarnings = [
 
 [tool.ruff]
 target-version = "py310"
-
+extend-exclude = [
+    "*/migrations/*.py",
+    ]
```
HawkiesZA, line 133: I just realised now that we exclude migration files from MyPy and Black (and with this PR, Ruff), but we do sometimes hand-write migrations (e.g. for a data migration). I know it's inconvenient to have to "fix" generated code, but shouldn't that code adhere to the same standard as the rest of the codebase?

fritzbrand: Mmm, good point.  @jerith your thoughts here?

home/tests/test_models.py
fritzbrand: Agreed, removed

home/whatsapp.py
fritzbrand: Jeremy and I worked on this together and cleaned things up.  We don't check specifically for a TypeError anymore, as we don't really care, as long as we can parse the json.  Please relook and let me know your thoughts

home/models.py
fritzbrand: Have added TODO to replace with PyParsing as discussed

home/models.py
fritzbrand: Have added TODO to replace with PyParsing as discussed

pyproject.toml
jerith: The main reason for consistent formatting and such is to make it easier to work on the code later. Since migrations need to remain unmodified once they've been applied, that doesn't really apply. Also, migration are also typically fairly small and self-contained.

We should still keep the same quality standard, but the noise of reformatting and effort of adding type annotations or whatever isn't worth the benefit we'd get from using those tools. There may be some benefit to running `ruff check` on migrations (possibly with some rules disabled), though.

home/models.py
```diff
@@ -1808,18 +1815,33 @@ def save_revision(
...
             )
-        except TemplateSubmissionException as e:
-            # The submission failed
+        except TemplateSubmissionServerException:
             revision.content["name"] = self.name
             revision.content["submission_name"] = self.template_name
             revision.content["submission_status"] = self.SubmissionStatus.FAILED
             revision.content["submission_result"] = (
-                f"Error! {e.response_json['error']['error_user_msg']} "
+                "An Internal Server Error has occurred.  Please try again later or contact developer support"
```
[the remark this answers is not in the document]
fritzbrand: @rudigiesler *Facepalm* of course, thanks.  Added

rudigiesler (approved):

PR #445, opened 2025-04-22
Resubmit standalone templates when button labels change

HawkiesZA:
## Purpose
Currently we don't resubmit standalone templates when the button titles change.

## Solution
This fixes the problem so that the templates resubmit.

#### Checklist
- [x] Added to release notes

### DIFF
--- a/home/models.py
+++ b/home/models.py
@@ -1702,7 +1702,6 @@ class WhatsAppTemplate(
         if previous_revision:
             previous_revision = previous_revision.as_object()
             previous_revision_fields = previous_revision.fields
-            print(f"Previous revision fields: {previous_revision_fields}")
             previous_revision_button_titles = [
                 b["value"]["title"] for b in previous_revision.buttons.raw_data
             ]
--- a/home/models.py
+++ b/home/models.py
@@ -861,7 +861,7 @@ class ContentPage(UniqueSlugMixin, Page, ContentImportMixin):
             self.whatsapp_template_category,
         )
 
-    def submit_whatsapp_template(self, previous_revision):
+    def submit_whatsapp_template(self, previous_revision: Revision) -> str | None:
         """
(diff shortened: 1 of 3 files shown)


home/models.py
```diff
@@ -1702,10 +1702,18 @@ def save_revision(
...
+            ]
         else:
             previous_revision_fields = ()
+            previous_revision_button_titles = []
 
-        if self.fields == previous_revision_fields:
+        current_button_titles = [b["value"]["title"] for b in self.buttons.raw_data]
+        if (
+            self.fields == previous_revision_fields
+            and current_button_titles == previous_revision_button_titles
```
rudigiesler, line 1715: Instead of adding an additional check here, can we not add a list of button titles to the `fields` property?

home/tests/test_models.py
```diff
@@ -1027,6 +1027,50 @@ def test_template_create_with_buttons(self) -> None:
...
+        wat.buttons = [
+            ("next_message", {"title": "Test Button 2"}),
+        ]
+        wat.save_revision().publish()
+
+        wat_from_db = WhatsAppTemplate.objects.last()
+
+        assert wat_from_db.submission_status == "SUBMITTED"
+        assert "Success! Template ID " in wat_from_db.submission_result
+        assert len(responses.calls) == 2
```
rudigiesler, line 1057: Can we also assert that the number of calls is 1 before changing the buttons? Just to make sure that there's nothing odd happening that's causing 2 calls on the first save, and none on the second?

home/models.py
HawkiesZA: hehe, saw this just before. Didn't realise the fields property existed. Much better.

home/tests/test_models.py
HawkiesZA: Good idea.

rudigiesler (approved):

PR #446, opened 2025-04-23
Improve variable validation

fritzbrand:
## Purpose
The validation of template variables have become overly convoluted, making it difficult to maintain and update

## Solution
We will implement PyParsing and use that to parse the variables first, and then use that list of parsed variables for validation

#### Checklist
- [ *] Added or updated unit tests
- [x] Added to release notes


home/tests/test_models.py
```diff
@@ -895,23 +895,64 @@ def test_clean_text_char_limit(self) -> None:
 
 @pytest.mark.django_db
 class TestWhatsAppTemplate:
-    def test_variables_are_numeric(self) -> None:
+    @override_settings(WHATSAPP_ALLOW_NAMED_VARIABLES=False)
+    def test_positional_variables_are_numeric(self) -> None:
         """
         Template variables are numeric.
         """
```
jerith, line 902: This tests that non-numeric positional variable are invalid, not that numeric ones are valid. Can you update the name and docstring, please?

home/tests/test_models.py
```diff
@@ -895,23 +895,64 @@ def test_clean_text_char_limit(self) -> None:
...
+            "message": [
+                "ParseException: Named variable can only contain alphanumberic and underscore characters. You provided '1_ok_not.ok'"
+            ],
+        }
+
+    @override_settings(WHATSAPP_ALLOW_NAMED_VARIABLES=True)
+    def test_int_and_named_variables_are_valid(self) -> None:
+        """
+        Named template variables must be alphanumeric or underscore
+        """
```
jerith, line 942: This test also needs a rename.

home/tests/test_models.py
```diff
@@ -1061,6 +1102,50 @@ def test_template_create_with_example_values(self) -> None:
...
+                    ("example_values", "Ev1"),
+                    ("example_values", "Ev2"),
+                ],
+            ).full_clean()
+
+        assert err_info.value.message_dict == {
+            "message": [
+                "Mismatch in number of placeholders and example values. Found 1 placeholder(s) and 2 example values."
+            ]
+        }
```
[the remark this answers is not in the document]
jerith: If named variables are allowed to be entirely numeric, we need a test to confirm that. (And also tests to consider various edge cases, like a numeric named parameter first in the list.)

home/whatsapp.py
```diff
@@ -375,3 +377,71 @@ def create_standalone_whatsapp_template(
     )
 
     return submit_whatsapp_template(name, category, locale, components)
+
+
+# Define parser grammar here
+vstart = pp.Literal("{{").suppress()
+vend = pp.Literal("}}").suppress()
+nonvar = pp.CharsNotIn("{}").suppress()
+if settings.WHATSAPP_ALLOW_NAMED_VARIABLES:
```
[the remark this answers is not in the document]
jerith: We also need to handle positional variables when named variables are allowed, so it would probably be cleaner for the parser to always accept named variables and then worry about named vs positional once we have the list of variables.

home/models.py
```diff
@@ -1766,6 +1754,12 @@ def clean(self) -> None:
                 ValidationError("All WhatsApp templates need a name.")
             )
 
+        message = self.message
+        try:
+            validate_template_variables(message)
```
jerith, line 1759: `validate_template_variables()` already returns a list of variables, so let's use that here instead of collecting them again with a regex below.

home/tests/test_whatsapp_standalone_templates.py
jerith, line 1: Since this is where we're doing the template variable validation, this is where we should test all the various edge cases. In the other places (models and import) we can test the success cases for positional and named variables, and one or two validation failures, then point to the tests here for more comprehensive coverage.

home/tests/test_whatsapp_template_import_export.py
jerith, line 1: I think we need one import test for named parameters, one for positional parameters, and one for invalid parameters. The docstring for the invalid parameter test can point to `test_whatsapp_standalone_templates.py` for more comprehensive coverage.

home/whatsapp.py
```diff
@@ -375,3 +377,71 @@ def create_standalone_whatsapp_template(
...
+            raise TemplateVariableError(
+                f"ParseException: Unable to parse the variable starting at character {pe.loc}"
+            )
+
+    if not variables:
+        return []
+
+    # If all the variables are ints, validate as positional variables
+    if all(
+        isinstance(var, int) or isinstance(var, str) and var.isdigit()
```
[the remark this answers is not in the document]
jerith: Perhaps `.isdecimal()` instead of `.isdigit()`?

home/tests/test_whatsapp_template_import_export.py
fritzbrand: Added a TODO as discussed, to deal with this once named vars are actually supported

home/tests/test_models.py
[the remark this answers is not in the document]
fritzbrand: - I have added a "happy path" test for named vars
- I don't think we should be going to deep into validating all the edge cases for named variables at this stage, and rather do that once we know all the rules around it (including what Turn will support)

home/whatsapp.py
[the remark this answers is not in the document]
fritzbrand: Have updated

home/tests/test_models.py
fritzbrand: Updated

home/tests/test_models.py
fritzbrand: Updated

home/models.py
fritzbrand: Done

rudigiesler (approved):

PR #447, opened 2025-04-24
Separate template submission from saving

HawkiesZA:
## Purpose
At the moment, whenever you save a new revision of a template, it gets submitted.

We want to separate the submission action from the saving action, so that we get to decide when it is submitted to Meta.

## Solution
This separates the submission action from the saving action and adds a submission button and a bulk submit action

#### Checklist
- [x] Added to release notes


home/models.py
```diff
@@ -939,14 +899,14 @@ def get_all_links(self):
 
     def save_revision(
```
rudigiesler, line 900: Can we just remove this method completely? It seems like we're just calling `super` and returning?

home/models.py
```diff
@@ -1696,50 +1643,6 @@ def save_revision(
...
-                "An Internal Server Error has occurred.  Please try again later or contact developer support"
-            )
-        except TemplateSubmissionClientException as tsce:
-            revision.content["name"] = self.name
-            revision.content["submission_name"] = self.template_name
-            revision.content["submission_status"] = self.SubmissionStatus.FAILED
-            revision.content["submission_result"] = str(tsce)
-
-        revision.save(update_fields=["content"])
         return revision
```
rudigiesler, line 1646: Same for this `save_revision`

home/whatsapp.py
```diff
@@ -346,6 +346,47 @@ def create_standalone_template_header_components(
     return components
 
 
+def submit_to_meta_menu_action(model: Any) -> None:
+    from .models import WhatsAppTemplate
+
+    revision = model.get_latest_revision()
```
rudigiesler, line 352: This might be wrong, so we should check it, but I think in the list view where you do the bulk action, it shows the latest published version, but when you go in and look at the edit view, it shows the latest revision.

So if that's true, for bulk action, should we not be submitting the latest published version? And for the individual submission, we submit the latest revision?

home/whatsapp.py
```diff
@@ -346,6 +346,47 @@ def create_standalone_template_header_components(
...
+        logger.exception(f"TemplateSubmissionServerException: {str(tsse)} ")
+        revision_object.submission_name = template_name
+        revision_object.submission_status = revision_object.SubmissionStatus.FAILED
+        revision_object.submission_result = "An Internal Server Error has occurred.  Please try again later or contact developer support"
+    except TemplateSubmissionClientException as tsce:
+        revision_object.submission_name = template_name
+        revision_object.submission_status = revision_object.SubmissionStatus.FAILED
+        revision_object.submission_result = str(tsce)
+
+    revision_object.save()
```
rudigiesler, line 387: Does this work? Or does this save to the live/published template? Do we not need to update `revision` instead, and then call `revision.save()` to overwrite the existing revision?

rudigiesler (commented):
Looks good!

I don't see where we're doing deduplication now though. If we submit the same template under a different name, Meta rejects it. I think since we're no longer submitting on every revision, we should look up the latest revision that has been submitted, and compare it to the current one. If it's changed, then we resubmit. If it hasn't, then we just copy the template name over to the current revision.

home/whatsapp.py
HawkiesZA: I've changed the method so that it takes in a WhatsAppTemplate. Whether that's the latest draft of live revision is then up to the calling function. Then at the end we save over the given object (i.e. submitting doesn't create a new revision, it just updates whatever you gave it)

rudigiesler: What I'm wondering here is, if you do `revision.as_object()`, you get a `WhatsAppTemplate`. If you then call `.save()` on that, does it then save to the `WhatsAppTemplate`, overwriting what's currently published? Since it's a WhatsAppTemplate object, not a revision? I don't know what happens if you call `.save()` on `revision.as_object()`, but maybe we should have a test that submitting a new revision that hasn't been published yet doesn't make it live.

rudigiesler (approved):

PR #448, opened 2025-04-25
Fix WA template export reimport

HawkiesZA:
## Purpose
Exporting and importing a WA template leads to an error on the import screen.

## Solution
The test was only checking if purge was set to true, which would clear the db and reimport. If purge is false though then the reimport fails due to there already being an entry. We now check if there is an entry and return it with the new data.

#### Checklist
- [ ] Added to release notes

### DIFF
--- a/home/import_whatsapp_templates.py
+++ b/home/import_whatsapp_templates.py
@@ -110,7 +110,20 @@ class WhatsAppTemplateImporter:
         self, row: "ContentRow", locale: Locale
     ) -> WhatsAppTemplate:
         try:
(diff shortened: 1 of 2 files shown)


home/import_whatsapp_templates.py
```diff
@@ -119,6 +106,40 @@ def create_whatsapp_template_from_row(self, row: "ContentRow") -> None:
         template.save()
         return
 
+    def _get_or_init_whatsapp_template(
```
rudigiesler, line 109: I think this is more `_update_or_create`, since it either fetches the existing one and updates it, or creates a new one and returns it.

rudigiesler (approved):

PR #457, opened 2025-05-08
Standalone template api

fritzbrand:
## Purpose
We are moving to a new version of the API, to expose the new Standalone Whatsapp Templates.  We need a temporary way to serve the old content page based templates under a V2 API, as well as the new standalone templates on a V3 API

## Solution
Implement the necessary serialisers and views to serve the V3 API, without breaking the existing V2 API

#### Checklist
- [ ] Added to release notes


home/api.py
```diff
@@ -121,6 +123,102 @@ def get_queryset(self):
...
+        ]
+    )
+    pagination_class = PageNumberPagination
+
+    def detail_view(self, request, pk):
+        try:
+            if "qa" in request.GET and request.GET["qa"] == "True":
+                instance = ContentPage.objects.get(
+                    id=pk
+                ).get_latest_revision_as_object()
```
HawkiesZA, line 150: I think this might be what we were talking about yesterday, where if we make a page, save it, but don't publish, then that page doesn't have a revision yet. I see we deal with this in get_queryset like this

```
queryset = ContentPage.objects.live().prefetch_related("locale")

if qa:
    queryset = queryset | ContentPage.objects.not_live()
```

home/api.py
```diff
@@ -121,6 +123,102 @@ def get_queryset(self):
...
+        # If this request is flagged as QA then we should display the pages that have the filtering tags
+        # or triggers in their draft versions
+        if "qa" in request.GET and request.GET["qa"] == "True":
+            tag = self.request.query_params.get("tag")
+            trigger = self.request.query_params.get("trigger")
+            have_new_triggers = []
+            have_new_tags = []
+            unpublished = ContentPage.objects.filter(has_unpublished_changes="True")
+            for page in unpublished:
+                latest_rev = page.get_latest_revision_as_object()
```
HawkiesZA, line 170: Similar here. If we have a page that hasn't been published it won't have any revisions.

home/api.py
```diff
@@ -121,6 +123,102 @@ def get_queryset(self):
...
+                if trigger and latest_rev.triggers.filter(name=trigger).exists():
+                    have_new_triggers.append(page.id)
+                if tag and latest_rev.tags.filter(name=tag).exists():
+                    have_new_tags.append(page.id)
+
+            queryset = self.get_queryset()
+            self.check_query_parameters(queryset)
+            queryset = self.filter_queryset(queryset)
+            queryset = queryset | ContentPage.objects.filter(id__in=have_new_triggers)
+            queryset = queryset | ContentPage.objects.filter(id__in=have_new_tags)
```
HawkiesZA, line 180: Do we only filter by either triggers or tags? Because otherwise I think it's possible to get duplicates in the final queryset here.

home/api.py
```diff
@@ -121,6 +123,102 @@ def get_queryset(self):
...
+        # or triggers in their draft versions
+        if "qa" in request.GET and request.GET["qa"] == "True":
+            tag = self.request.query_params.get("tag")
+            trigger = self.request.query_params.get("trigger")
+            have_new_triggers = []
+            have_new_tags = []
+            unpublished = ContentPage.objects.filter(has_unpublished_changes="True")
+            for page in unpublished:
+                latest_rev = page.get_latest_revision_as_object()
+                if trigger and latest_rev.triggers.filter(name=trigger).exists():
```
HawkiesZA, line 171: What happens if we supply an invalid trigger or tag?

home/serializers.py
```diff
@@ -626,3 +842,73 @@ def to_representation(self, page):
...
+        return {
+            "id": template.id,
+            "locale": template.locale.language_code,
+            "name": name_field_representation(template, request),
+            "category": category_field_representation(template, request),
+            "image": template.image,
+            "message": template.message,
+            "example_values": list(template.example_values.raw_data),
+            "buttons": list(template.buttons.raw_data),
+            "revision": template.get_latest_revision().id,
```
HawkiesZA, line 908: What happens if the template doesn't have a revision?

home/serializers.py
```diff
@@ -320,6 +375,143 @@ def body_field_representation(page, request):
...
+        except ValueError:
+            raise ValidationError(
+                "Please insert a positive integer for message in " "the query string"
+            )
+    else:
+        message = 0
+
+    if "whatsapp" in request.GET and (
+        page.enable_whatsapp is True
+        or ("qa" in request.GET and request.GET["qa"] == "True")
```
HawkiesZA, line 394: Same possible confusion here with lowercase :-)

home/serializers.py
```diff
@@ -320,6 +375,143 @@ def body_field_representation(page, request):
...
+                            (
+                                "previous_message",
+                                has_previous_message(message, page, "whatsapp"),
+                            ),
+                            ("total_messages", len(page.whatsapp_body._raw_data)),
+                            (
+                                "message",
+                                whatsapp_messages[message],
+                            ),
+                            ("revision", page.get_latest_revision().id),
```
HawkiesZA, line 417: Looks like "message" is in here twice? I think "message" is an id, and maybe "text" for the message text.

fritzbrand: I think this bit of code was removed as per a comment from Rudi

home/serializers.py
```diff
@@ -626,3 +842,73 @@ def to_representation(self, page):
...
+
+def name_field_representation(template, request):
+    return template.name
+
+
+class WhatsAppTemplateSerializer(BaseSerializer):
+    # name = NameField(read_only=True)
+    # message = MessageField(read_only=True)
+    # category = CategoryField(read_only=True)
+    # TODO: Not sure which fields should or not be def'd up here, check with clever people
```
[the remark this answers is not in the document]
fritzbrand: Done

home/serializers.py
```diff
@@ -320,6 +375,143 @@ def body_field_representation(page, request):
...
+        message = 0
+
+    if "whatsapp" in request.GET and (
+        page.enable_whatsapp is True
+        or ("qa" in request.GET and request.GET["qa"] == "True")
+    ):
+        if page.whatsapp_body != []:
+            whatsapp_messages = format_whatsapp_body(message, page)
+
+            if specific_message:
```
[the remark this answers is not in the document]
fritzbrand: Nah agreed, removed this

home/api.py
fritzbrand: Moved to new ticket

home/api.py
fritzbrand: Moved to new ticket

home/serializers.py
```diff
@@ -665,3 +845,86 @@ def to_representation(self, page):
...
+class MessageField(serializers.Field):
+    """
+    Serializes the "message" field.
+    """
+
+    def get_attribute(self, instance):
+        return instance
+
+    def to_representation(self, instance):
+        return instance.message
```
rudigiesler, line 871: These all seem like quite simple serializations. Could we not just use something like a `CharField`, or does the base ModelSerializer not just do this correctly for us, and we can just specify them in the `fields` list?

home/serializers.py
```diff
@@ -665,3 +845,86 @@ def to_representation(self, page):
...
+    def to_representation(self, instance):
+        return instance.category
+
+
+def category_field_representation(template, request):
+    return template.category
+
+
+def name_field_representation(template, request):
+    return template.name
```
rudigiesler, line 891: It doesn't seem like we're using these functions anywhere

home/serializers.py
```diff
@@ -359,6 +415,106 @@ def body_field_representation(page: Any, request: Any) -> Any:
...
+                        ),
+                        ("total_messages", len(page.messenger_body._raw_data)),
+                        ("text", page.messenger_body._raw_data[message]["value"]),
+                    ]
+                )
+            except IndexError:
+                raise ValidationError("The requested message does not exist")
+    elif "viber" in request.GET and (
+        page.enable_viber is True
+        or ("qa" in request.GET and request.GET["qa"] == "True")
```
rudigiesler, line 498: I don't think that this is correct logic, if `enable_viber` is `False` in the latest revision, then we should not show it, but this logic will. I think this whole section needs a refactor, but if we're doing that in the next ticket, then I'm okay with this going through in this one.

home/tests/test_api_v3.py
```diff
@@ -0,0 +1,1026 @@
...
+
+    #     assert item_1 == {"type": "next_message", "value": {"title": "list item 1"}}
+    #     assert item_2 == {
+    #         "type": "go_to_page",
+    #         "value": {"title": "list item 2", "page": target_page.id},
+    #     }
+    #     assert item_3 == {
+    #         "type": "go_to_form",
+    #         "value": {"title": "list item 3", "form": form.id},
+    #     }
```
rudigiesler, line 1026: There are a lot of full integration tests here, but I think we can simplify the testing a lot of we have one or two integration tests, and then a whole bunch of unit tests, testing things like the edge cases in the serializer logic. But since most of this is commented out, I'm assuming that this is something that will be looked at in the next ticket, so happy for that to take place there.

home/serializers.py
```diff
@@ -428,6 +584,30 @@ def to_representation(self, page):
...
+            "meta": metadata_field_representation(page, request, router),
+            "title": title_field_representation(page, request),
+            "subtitle": subtitle_field_representation(page),
+            "messages": body_field_representation_v3(page, request),
+            "tags": [x.name for x in page.tags.all()],
+            "triggers": [x.name for x in page.triggers.all()],
+            "quick_replies": [x.name for x in page.quick_replies.all()],
+            "has_children": has_children_field_representation(page),
+            "related_pages": related_pages_field_representation(page, request),
+        }
```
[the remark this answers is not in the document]
fritzbrand: Left a TODO: comment for you in the code

home/migrations/0102_remove_whatsapptemplate_name_whatsapptemplate_slug.py
```diff
@@ -0,0 +1,37 @@
...
+                max_length=255,
+            ),
+        ),
+       
+        migrations.AddConstraint(
+            model_name="whatsapptemplate",
+            constraint=models.UniqueConstraint(
+                fields=("slug", "locale"), name="unique_slug_locale"
+            ),
+        ),
```
rudigiesler, line 36: I think this migration will fail. Since we're removing and then adding a field with a default value of "", that means that all slugs will just be empty strings, and so there will definitely be duplicates.

What we need to do instead is first add the slug field, make the default value a slugified version of the name field, and then remove the name field, and then add the constraint.

home/serializers.py
```diff
@@ -224,43 +227,30 @@ def body_field_representation(page: Any, request: Any) -> Any:
                                 format_whatsapp_template_message(template.message),
                             ),
                             ("is_whatsapp_template", "True"),
-                            ("whatsapp_template_name", template.name),
+                            ("whatsapp_template_slug", template.slug),
```
rudigiesler, line 230: Doesn't this mean that we're changing the v2 API? Shouldn't we keep it `whatsapp_template_name`, and just have it reference `template.slug`?

home/serializers_v3.py
```diff
@@ -0,0 +1,433 @@
...
+    def to_representation(self, page):
+        request = self.context["request"]
+        return body_field_representation_V3(page, request)
+
+
+def body_field_representation_V3(page, request):
+    message = 0
+    if "whatsapp" in request.GET and (
+        page.enable_whatsapp is True
+        or ("qa" in request.GET and request.GET["qa"] == "True")
```
rudigiesler, line 121: I think this is another place with the query params where things are wrong that can be solved in the next PR. This checks if whatsapp is enabled or if we've asked for QA. Instead, if it's QA, we should be checking if the latest revision has whatsapp enabled (ideally we would just pass the latest revision to this function, and this function doesn't need to worry about QA)

home/serializers_v3.py
```diff
@@ -0,0 +1,433 @@
...
+        request = self.context["request"]
+        messages = body_field_representation_V3(obj, request)
+        return messages
+
+    def get_buttons(self, obj):
+        buttons = list(obj.buttons.raw_data)
+        for button in buttons:
+            if "id" in button:
+                del button["id"]
+        return buttons
```
rudigiesler, line 376: Can we not use the `format_buttons_and_list_items` function here? Also, I don't think we do buttons or example values directly here, so we can probably remove these two functions? (they're not in the fields list)

fritzbrand: Done

home/serializers_v3.py
fritzbrand: Yes, will tackle in next PR

home/migrations/0102_remove_whatsapptemplate_name_whatsapptemplate_slug.py
fritzbrand: Cool, have updated

home/api_v3.py
```diff
@@ -0,0 +1,251 @@
...
+        elif "whatsapp" in self.request.query_params:
+            queryset = queryset.filter(enable_whatsapp=True)
+        elif "sms" in self.request.query_params:
+            queryset = queryset.filter(enable_sms=True)
+        elif "ussd" in self.request.query_params:
+            queryset = queryset.filter(enable_ussd=True)
+        elif "messenger" in self.request.query_params:
+            queryset = queryset.filter(enable_messenger=True)
+        elif "viber" in self.request.query_params:
+            queryset = queryset.filter(enable_viber=True)
```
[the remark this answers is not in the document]
fritzbrand: Was also planning on doing this with my refactoring of the query param filtering stuff (next pr)

home/api_v3.py
```diff
@@ -0,0 +1,251 @@
...
+                # )
+                # return Response(serializer.data)
+            else:
+                # WhatsAppTemplate.objects.get(id=pk)
+                instance = self.get_object()
+        except Exception as e:
+            # TODO: Handle this better
+            print(f"Exception = {e}")
+
+        serializer = WhatsAppTemplateSerializer(instance, context={"request": request})
```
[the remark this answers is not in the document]
rudigiesler: If you couldn't get it working that way, probably not worth the effort, probably some wagtail override thing making it difficult and act differently to normal DRF things, happy with the way it is :)

rudigiesler (approved):

PR #459, opened 2025-05-12
Fix being able to submit to meta after import

HawkiesZA:
## Purpose
After importing a standard template, submitting to Meta throws an exception until the template is saved and published.

## Solution
We create the first revision on import.

#### Checklist
- [ ] Added to release notes

### DIFF
--- a/home/tests/test_whatsapp_template_import_export.py
+++ b/home/tests/test_whatsapp_template_import_export.py
@@ -141,10 +141,39 @@ def remove_button_ids(template: DbDict) -> DbDict:
     return template
 
 
+PAGE_TIMESTAMP_FIELDS = {
+    "first_published_at",
+    "last_published_at",
+    "latest_revision_created_at",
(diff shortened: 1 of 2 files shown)


home/tests/test_whatsapp_template_import_export.py
```diff
@@ -141,10 +141,39 @@ def remove_button_ids(template: DbDict) -> DbDict:
...
+    templates = _update_field(templates, "live_revision", lambda v: v - min_live)
+    return templates
+
+
 WHATSAPP_TEMPLATE_FILTER_FUNCS = [
     remove_example_value_ids,
     normalise_pks,
     remove_button_ids,
+    remove_timestamps,
+    # normalise_revisions,
```
rudigiesler, line 176: If we're not using it, we should remove it... But I think we would rather normalise than delete if possible (which is the current test implementation)

PR #461, opened 2025-05-13
Use the page name instead of what could be an empty field

HawkiesZA:
## Purpose
We discovered that the whatsapp_template_name field can be empty for unsubmitted templates, and that's not what we want anyway.

## Solution
Use the code from the prefix.

#### Checklist
- [ ] Added to release notes

### DIFF
--- a/home/migrations/0099_migrate_content_page_templates_to_standalone_templates.py
+++ b/home/migrations/0099_migrate_content_page_templates_to_standalone_templates.py
@@ -1,5 +1,45 @@
 from django.db import migrations
 from typing import Any
+from django.db.models import Count
+
+
+def rename_duplicate_whatsapp_titles(ContentPage: Any) -> None:
+    duplicate_whatsapp_titles = (
+        ContentPage.objects.values("whatsapp_title")
+        .annotate(count=Count("whatsapp_title"))
+        .filter(count__gt=1)
+        .values_list("whatsapp_title", flat=True)
+    )
+    for whatsapp_title in duplicate_whatsapp_titles:
+        pages = ContentPage.objects.filter(whatsapp_title=whatsapp_title)
+        while pages.count() > 1:
+            page = pages.first()
+            suffix = 1
+            candidate_whatsapp_title = whatsapp_title
+            while ContentPage.objects.filter(
+                whatsapp_title=candidate_whatsapp_title
+            ).exists():
+                suffix += 1
+                candidate_whatsapp_title = f"{whatsapp_title}-{suffix}"
+            page.whatsapp_title = candidate_whatsapp_title
+            page.save(update_fields=["whatsapp_title"])
+
+
+def rename_matching_whatsapp_names(ContentPage: Any, WhatsAppTemplate: Any) -> None:
+    templates = WhatsAppTemplate.objects.all()
+    for template in templates:
+        pages = ContentPage.objects.filter(whatsapp_title=template.name)
+        while pages.count() > 1:
+            page = pages.first()
+            suffix = 1
+            candidate_whatsapp_title = template.name
+            while ContentPage.objects.filter(
+                whatsapp_title=candidate_whatsapp_title
+            ).exists():
+                suffix += 1
+                candidate_whatsapp_title = f"{template.name}-{suffix}"
+            page.whatsapp_title = candidate_whatsapp_title
+            page.save(update_fields=["whatsapp_title"])
 
 
 def migrate_content_page_templates_to_standalone_templates(
@@ -32,7 +72,7 @@ def migrate_content_page_templates_to_standalone_templates(
                 print("No image")
             example_values = list(whatsapp_value.get("example_values", []))
             whatsapp_template = WhatsAppTemplate.objects.create(
-                name=f'{content_page.whatsapp_title.lower().replace(" ", "_")}_{content_page.get_latest_revision().pk}',
+                name=content_page.whatsapp_title.lower().replace(" ", "_"),
                 locale=content_page.locale,
                 message=whatsapp_value.get("message", ""),
                 example_values=[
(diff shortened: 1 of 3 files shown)


jerith (approved):

home/migrations/0099_migrate_content_page_templates_to_standalone_templates.py
```diff
@@ -1,37 +1,105 @@
...
 from typing import Any
+from django.db.models import Count
+
+
+def rename_duplicate_whatsapp_titles(ContentPage: Any) -> None:
+    duplicate_whatsapp_titles = (
+        ContentPage.objects.values("whatsapp_title")
+        .annotate(count=Count("whatsapp_title"))
+        .filter(count__gt=1)
+        .values_list("whatsapp_title", flat=True)
```
rudigiesler, line 11: We only need to do this for `ContentPage`s with `is_whatsapp_template=True` right?

home/migrations/0099_migrate_content_page_templates_to_standalone_templates.py
```diff
@@ -1,37 +1,105 @@
...
+                suffix += 1
+                candidate_whatsapp_title = f"{whatsapp_title}-{suffix}"
+            page.whatsapp_title = candidate_whatsapp_title
+            page.save(update_fields=["whatsapp_title"])
+
+
+def rename_matching_whatsapp_names(ContentPage: Any, WhatsAppTemplate: Any) -> None:
+    templates = WhatsAppTemplate.objects.all()
+    for template in templates:
+        pages = ContentPage.objects.filter(whatsapp_title=template.name)
```
rudigiesler, line 31: Same here, we probably only care for `is_whatsapp_template=True` content pages

home/migrations/0099_migrate_content_page_templates_to_standalone_templates.py
```diff
@@ -1,37 +1,105 @@
...
-            submission_status="NOT_SUBMITTED_YET",
-            submission_result="",
-        )
-        content_page.whatsapp_body = []
-        content_page.whatsapp_body.append(("Whatsapp_Template", whatsapp_template))
-        content_page.is_whatsapp_template = False
-        content_page.save()
+        print(f"Content Page slug: {content_page.slug}")
+        print(f"Content page title: {content_page.whatsapp_title}")
+        print(f"Whatsapp Value type: {type(whatsapp_value)}, value: {whatsapp_value}")
```
rudigiesler, line 55: I don't think we need these debugging prints any more. If we want to have some output, better to use the logging module.

home/migrations/0099_migrate_content_page_templates_to_standalone_templates.py
```diff
@@ -1,37 +1,105 @@
...
+        if (
+            not hasattr(whatsapp_value, "_meta")
+            or whatsapp_value._meta.verbose_name != "WhatsApp Template"
+        ):
```
rudigiesler, line 59: Could we not rather do the inverse of this, and use the `continue` keyword, so that we can avoid this deep nesting? e.g.
```python
if hasattr(whatsapp_value, "_meta") and whatsapp_value._meta.verbose_name == "WhatsApp Template":
    continue
```

home/migrations/0099_migrate_content_page_templates_to_standalone_templates.py
```diff
@@ -1,37 +1,105 @@
...
+                    image = Image.objects.get(id=image.id)
+                elif isinstance(image, dict) and "id" in image:
+                    image = Image.objects.get(id=image["id"])
+                else:
+                    image = None
+            else:
+                print("No image")
+            example_values = list(whatsapp_value.get("example_values", []))
+            whatsapp_template = WhatsAppTemplate.objects.create(
+                name=content_page.whatsapp_title.lower().replace(" ", "_"),
```
rudigiesler, line 75: Instead of the two previous functions, a more robust way would be to have a function here to create a template name, which would do the looping to create a unique template name, and then we pass that value to `WhatsAppTemplate.objects.create`

home/migrations/0099_migrate_content_page_templates_to_standalone_templates.py
```diff
@@ -1,37 +1,105 @@
...
+                    ("example_values", example_value)
+                    for example_value in example_values
+                ],
+                category=content_page.whatsapp_template_category,
+                buttons=whatsapp_value.get("buttons", []),
+                image=image,
+                submission_status="NOT_SUBMITTED_YET",
+                submission_result="",
+            )
+            content_page.whatsapp_body = []
```
rudigiesler, line 88: This will replace all the messages for the content page will it not? If the content page has multiple messages, then we'll want to replace the first one with the template, and leave the rest of them as normal messages, not get rid of them

rudigiesler (commented):
To add to the comments, I think we'll want to add some more migration tests to cover these new edge cases.

home/migrations/0099_migrate_content_page_templates_to_standalone_templates.py
HawkiesZA: Wasn't sure if the logging module would do the job on QA, Jeremy suggested to just use prints :-) Will take them out anyway.

home/migrations/0099_migrate_content_page_templates_to_standalone_templates.py
HawkiesZA: Ok, I went back in time so that the test will work, adjusted the code and checked the test. I also then trashed my local DB, and recreated it up until just before this migration, created a template with 2 messages, and ran the migration. Result: I got a new Standalone Template and a Page with 2 messages - the first one was the template and the second was the second message as before.

home/migrations/0099_migrate_content_page_templates_to_standalone_templates.py
```diff
@@ -1,37 +1,102 @@
...
+            elif hasattr(image, "id"):
+                image = Image.objects.get(id=image.id)
+            elif isinstance(image, dict) and "id" in image:
+                image = Image.objects.get(id=image["id"])
+            else:
+                image = None
+        example_values = list(whatsapp_value.get("example_values", []))
         whatsapp_template = WhatsAppTemplate.objects.create(
-            name=content_page.whatsapp_template_name,
+            name=content_page.whatsapp_title.lower().replace(" ", "_"),
```
rudigiesler, line 73: We could still have conflicts of template names because of this, `Template 1` and `template_1` won't be caught in the `rename` functions, but we'll get an integrity error here.

I think instead of trying to prep the data in the `rename` functions, we should instead create a candidate name here, and make it unique against all existing whatsapp templates, before saving it. Then we don't need to worry about where it's coming from, we just take it one template at a time, creating unique names.

rudigiesler (approved):

PR #462, opened 2025-05-21
Add the locale into the template submission name

HawkiesZA:
Add the locale into the template submission name to prevent duplicate submissions across locales

## Purpose
When submitting to meta it's possible to have duplicate submission names across locales.

## Solution
Include the locale in the submission name.

#### Checklist
- [ ] Added to release notes

### DIFF
--- a/home/models.py
+++ b/home/models.py
@@ -1589,7 +1589,7 @@ class WhatsAppTemplate(
 
     @property
     def prefix(self) -> str:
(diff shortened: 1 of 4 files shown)


rudigiesler (commented):
I don't understand why this is necessary. We set the `language` field when we submit the whatsapp template, and templates with the same `name` but different values for `language` are allowed by the whatsapp template creation API, that's how translations are handled.

PR #466, opened 2025-06-11
Fixed V2 view issue, and added tests

fritzbrand:
## Purpose
A bug was logged relating to the V2 API.  The detail page was not showing all the fields it should

## Solution
I messed up and forgot to complete the work required to make a new standalone template, look like a normal "old" whatsapp message when looked at via the V2 API.  This PR adds that work, including tests

### DIFF
--- a/home/tests/page_builder.py
+++ b/home/tests/page_builder.py
@@ -302,9 +302,6 @@ class PageBuilder(Generic[TPage]):
         triggers: Iterable[str] | None = None,
         quick_replies: Iterable[str] | None = None,
         buttons: Iterable[str] | None = None,
-        whatsapp_template_name: str | None = None,
-        whatsapp_template_slug: str | None = None,
-        whatsapp_template_category: str | None = None,
         translated_from: ContentPage | None = None,
         publish: bool = True,
     ) -> ContentPage:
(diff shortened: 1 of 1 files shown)


home/tests/test_api.py
```diff
@@ -858,8 +859,10 @@ def create_content_page(
         list_items: list[str] | None = None,
         next_prompt: str | None = None,
         footer: str | None = None,
+        whatsapp_template_example_values: list[ExVal] | None = None,
+        whatsapp_template_image: int | None = None,
         whatsapp_template_category: str | None = None,
-        # whatsapp_template_submission_name: str | None = None,
+        whatsapp_template_submission_name: str | None = None,
         whatsapp_template_slug: str | None = None,
```
jerith, line 866: Do these fields still exist on `ContentPage`?

home/tests/test_api.py
```diff
@@ -991,15 +1010,76 @@ def test_whatsapp_template_fields(self, uclient: Any) -> None:
...
+        page = self.create_content_page(
+            whatsapp_template_category=ContentPage.WhatsAppTemplateCategory.MARKETING,
+            whatsapp_template_slug="test_template",
+            whatsapp_template_image=1,
+            whatsapp_template_example_values=[
+                ("example_values", "Ev1"),
+                ("example_values", "Ev1"),
+            ],
+            whatsapp_template_submission_name="test_template_1",
+            buttons=[{"type": "next_message", "value": {"title": "Tell me more"}}],
```
jerith, line 1036: Do we need a similar test with list options instead of buttons? (Similarly for the non-template version, and also in the v3 api.)

home/tests/test_api.py
fritzbrand: Yes, we use them to create the template inside `create_content_page`

rudigiesler (approved):
I'm not sure why we're commenting out all of the v3 API tests, but otherwise looks good :)

PR #470, opened 2025-07-08
Refactor param filtering 4

fritzbrand:
## Purpose
During our work on the Standalone Templates API, we picked up some bugs in the parameter filtering code, and that the code was too complex and fragmented to do an easy fix.   

## Solution
We decided to refactor the code to run in memory, rather than making multiple queries at different stages

#### Checklist
- [ ] Added to release notes

### DIFF
--- a/home/api_v3.py
+++ b/home/api_v3.py
@@ -37,7 +37,7 @@ class WhatsAppTemplateViewset(BaseAPIViewSet):
             self.lookup_field = "slug"
 
         try:
-            if "qa" in request.GET and request.GET["qa"] == "True":
+            if "qa" in request.GET and request.GET["qa"].lower() == "true":
                 instance = self.get_object().get_latest_revision_as_object()
             else:
                 instance = self.get_object()
@@ -140,7 +140,7 @@ class ContentPagesV3APIViewset(PagesAPIViewSet):
     def listing_view(self, request, *args, **kwargs):
         # If this request is flagged as QA then we should display the pages that have the filtering tags
         # or triggers in their draft versions
-        if "qa" in request.GET and request.GET["qa"] == "True":
+        if "qa" in request.GET and request.GET["qa"].lower() == "true":
             tag = self.request.query_params.get("tag")
             trigger = self.request.query_params.get("trigger")
             have_new_triggers = []
--- a/home/api_v3.py
+++ b/home/api_v3.py
@@ -17,7 +17,10 @@ from .models import (  # isort:skip
     WhatsAppTemplate,
 )
 
-import pprint as pp
+import pprint
+
+pp = pprint.PrettyPrinter(indent=4)
+pp.pprint("Forget me not")
 
 
 class WhatsAppTemplateViewset(BaseAPIViewSet):
@@ -122,7 +125,10 @@ class ContentPagesV3APIViewset(PagesAPIViewSet):
             self.lookup_field = "pk"
 
         try:
-            if "qa" in request.GET and request.GET["qa"].lower() == "true":
+            if (
+                "return_drafts" in request.GET
+                and request.GET["return_drafts"].lower() == "true"
+            ):
                 instance = self.get_object().get_latest_revision_as_object()
             else:
                 instance = self.get_object()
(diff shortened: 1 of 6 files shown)


home/api_v3.py
```diff
@@ -123,7 +123,10 @@ def detail_view(self, request, pk=None, slug=None):
             self.lookup_field = "pk"
 
         try:
-            if "qa" in request.GET and request.GET["qa"].lower() == "true":
+            if (
+                "return_drafts" in request.GET
+                and request.GET["return_drafts"].lower() == "true"
```
jerith, line 128: Why not use `request.query_params.get("qa", "").lower()` like we do in v2?

home/api_v3.py
```diff
@@ -179,40 +160,77 @@ def listing_view(self, request, *args, **kwargs):
...
+
         trigger = self.request.query_params.get("trigger")
         if trigger is not None:
             ids = []
             for t in TriggeredContent.objects.filter(tag__name__iexact=trigger.strip()):
                 ids.append(t.content_object_id)
-            queryset = queryset.filter(id__in=ids)
+            live_queryset = live_queryset.filter(id__in=ids)
+            draft_queryset = draft_queryset.filter(id__in=ids)
+            combined_queryset = combined_queryset.filter(id__in=ids)
```
jerith, line 231: Do we really need to carry these all the way through?

home/api_v3.py
```diff
@@ -179,40 +160,77 @@ def listing_view(self, request, *args, **kwargs):
...
-        elif "messenger" in self.request.query_params:
-            queryset = queryset.filter(enable_messenger=True)
-        elif "viber" in self.request.query_params:
-            queryset = queryset.filter(enable_viber=True)
+        return_drafts = (
+            self.request.query_params.get("return_drafts", "").lower() == "true"
+        )
+        if return_drafts:
+            draft_queryset = draft_queryset | live_queryset
+            combined_queryset = draft_queryset | live_queryset
```
jerith, line 173: This looks like we're `|`ing `live_queryset` twice.

fritzbrand: I have reworked this section a bit, please have another look

home/api_v3.py
```diff
@@ -103,74 +104,58 @@ class ContentPagesV3APIViewset(PagesAPIViewSet):
...
             self.lookup_field = "slug"
+
         else:
             self.lookup_field = "pk"
 
         try:
-            if "qa" in request.GET and request.GET["qa"].lower() == "true":
+            if (
+                "return_drafts" in request.GET
+                and self.request.query_params.get("return_drafts", "").lower() == "true"
```
jerith, line 131: I don't think you need the `"return_drafts" in request.GET` anymore, because `query_params.get("return_drafts", "")` takes care of that.

home/api_v3.py
```diff
@@ -103,74 +104,58 @@ class ContentPagesV3APIViewset(PagesAPIViewSet):
...
+        channel = ""
+        if "channel" in self.request.query_params:
+            channel = self.request.query_params.get("channel", "").lower()
+
+            if channel in {"web", "whatsapp", "sms", "ussd", "messenger", "viber"}:
+                queryset = queryset.filter(**{f"enable_{channel}": True})
+            else:
+                raise ValidationError(
+                    {"channel": [f"Channel matching query '{channel}' does not exist."]}
+                )
```
jerith, line 157: I think we can factor out the channel lookup and validation check into a function that we call anywhere we need to get the channel:

```python
channel = get_channel(self.request.query_params)
if channel:
    queryset = queryset.filter(**{f"enable_{channel}": True})
```

home/api_v3.py
```diff
@@ -97,122 +132,212 @@ class ContentPagesV3APIViewset(PagesAPIViewSet):
...
+    def _build_filters(self) -> Q:
+        """Build query filters based on request parameters."""
+        base_filters = Q()
+
+        # Get and normalize query parameters
+        locale = self.request.query_params.get("locale", DEFAULT_LOCALE).casefold()
+        slug = self.request.query_params.get("slug", "").casefold()
+        title = self.request.query_params.get("title", "").casefold()
+        trigger = self.request.query_params.get("trigger", "").casefold()
+        tag = self.request.query_params.get("tag", "").casefold()
```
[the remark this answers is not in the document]
fritzbrand: Not sure if this still holds true, as we now also use these variables to check against the latest rev details (if return_draft is true), using casefold in those checks, so maybe it's best to leave this in place?

rudigiesler (commented):
Looks good! Just a single comment about some newlines that I think we can reduce, otherwise happy with it

PR #536, opened 2026-06-04
Use new tools

HawkiesZA:
## Purpose
We should be using uv, ty, and ruff as our main tools in our repos.

## Solution
This migrates from poetry to uv, black to ruff, mypy to ty.

#### Checklist
- [ ] Added to release notes


home/import_helpers.py
```diff
@@ -304,7 +304,7 @@ def read_xlsx(file_content: bytes) -> Iterator[dict[str, Any]]:
     header = remove_trailing_nones(header)
 
     for row in worksheet.iter_rows(min_row=2, values_only=True):
-        row = remove_trailing_nones(row)  # type: ignore # Mypy cannot guarantee row is a list; rows may be tuples so we bypass.
+        row = remove_trailing_nones(row)  # type: ignore # Static analysis cannot guarantee row is a list; rows may be tuples so we bypass.
```
jerith: Could we perhaps have `remove_trailing_nones` take a `Sequence` instead of a `list`?
(Happy for this to happen later, though.)

home/serializers.py
```diff
@@ -21,7 +21,8 @@ class TitleField(serializers.Field):
     def get_attribute(self, instance):
         return instance
 
-    def to_representation(self, page):
+    def to_representation(self, value):
+        page = value
         request = self.context["request"]
         return title_field_representation(page, request)
```
jerith: Is there a reason we assign to a new variable here instead of passing `value` to `title_field_representation`?

jerith (approved):

rudigiesler (approved):

